PR #1, opened 2020-08-26
Add core package with TS declarations

emmenko:
I've got it to a point where the types compile and the autocompletion is helpful.

image

image

There is still some cleanup to do, but the tests work (the same as we have internally) and this is a good sign :)

### DIFF
--- a/core/src/builder.ts
+++ b/core/src/builder.ts
@@ -4,6 +4,7 @@ import type {
   TBuilderOptions,
   TBuilder,
   TPropertyBuilder,
+  TTransformBuildName,
 } from './types';
 
 import {
@@ -11,7 +12,7 @@ import {
   isString,
   convertBuiltNameToTransformName,
   onlyProps,
-  removeProps,
+  omitMany,
 } from './helpers';
 
 interface CustomProxyConstructor {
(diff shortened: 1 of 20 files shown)


core/src/generator.ts
```diff
@@ -0,0 +1,79 @@
+import type { Overrides } from '@jackfranklin/test-data-bot';
+import type { TGeneratorOptions, TGeneratorResult } from './types';
+
+import { build } from '@jackfranklin/test-data-bot';
+
+// /* TYPES DECLARATIONS FROM @jackfranklin/test-data-bot */
```
emmenko, line 6: I opened a PR to export the types: https://github.com/jackfranklin/test-data-bot/pull/344

emmenko:
@tdeekens @adnasa this is ready for review, whenever you have time.

PS: if you need a higher level introduction about the implementation and types, let me know and we can schedule a meeting. Thanks

tdeekens:
One high level note. This is valid in regards to using and defining test-data.

```js
const preset = customer.random().id('foo').name('Tobi');

const nicola = preset.name('Nicola');
```

This implies that a prop of a builder can be set twice or overwritten. This should be allowed by the types too. How I see it we have put effort in not allowing this?

emmenko:
I see. So you can potentially do `Builder().name('1').name('2').name('3')`?

adnasa (approved):
I think this looks good. I like Tobi's suggestion, but I don't know if we want to do this on this PR or on a separate step.

core/src/builder.ts
```diff
@@ -0,0 +1,189 @@
...
+                  transformed = (transformers?.default?.transform(built) ??
+                    built) as Model;
+                  break;
+                }
+                case 'buildGraphql': {
+                  transformed = (transformers?.graphql?.transform(built) ??
+                    built) as Model;
+                  break;
+                }
+                case 'buildRest': {
```
tdeekens, line 134: Probably for the best.

tdeekens (commented):
Very nice. Nothing to add!

tdeekens (approved):

adnasa:
😍

PR #4, opened 2020-09-24
fix(core): transformer types, replaceFields option takes higher precedence now

emmenko:
We started implementing some models in appkit (https://github.com/commercetools/merchant-center-application-kit/pull/1743) and bumped into some small issues.
This PR fixes that.

### DIFF
--- a/core/src/builder.spec.ts
+++ b/core/src/builder.spec.ts
@@ -167,48 +167,12 @@ describe('building', () => {
       });
 
       describe('when replacing fields', () => {
-        describe('when field does not exist', () => {
-          it('should not add the desired fields', () => {
-            const transformers = {
-              graphql: Transformer<TestUser, TestUser>('graphql', {
-                // The `identifier` is not part of the `TestUser`, so TS yells at us as expected.
-                // However for the purpose of the test, we ignore the type check.
-                // @ts-ignore
-                replaceFields: ({ fields }) => ({
-                  identifier: fields.id,
-                }),
-              }),
-            };
-            const built = Builder<TestUser>({
-              transformers,
-            })
-              .id(1)
-              .userName('Fred')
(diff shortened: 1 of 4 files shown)


core/src/builder.spec.ts
```diff
@@ -167,15 +167,13 @@ describe('building', () => {
...
+          it('should replace the desired fields', () => {
             const transformers = {
               graphql: Transformer<TestUser, TestUser>('graphql', {
-                // The `identifier` is not part of the `TestUser`, so TS yells at us as expected.
-                // However for the purpose of the test, we ignore the type check.
-                // @ts-ignore
                 replaceFields: ({ fields }) => ({
-                  identifier: fields.id,
+                  ...fields,
+                  id: fields.id + 1,
```
emmenko, line 176: The `replaceFields` now should return the fully final transformed object, not just a partial representation.

This is important because the result of a transform does not necessarily need to be an object, but can also be an array, a string, a number, etc.

tdeekens: Sorry, why can't I just mutate? Cause otherwise I can't mutate primitives?

emmenko: What do you mean by "mutate"?

tdeekens: Instead of returning something new just mutating it. `field.id = field.id + 1` then `return field`?

emmenko: Ah you mean just in this test example?

tdeekens: I think I just misunderstood the initial concern. It's fine that the whole field has to be returned not just a partial of it. If it's then being mutated or if it's an shallowly immutable clone doesn't matter.

tdeekens (approved):

adnasa (approved):

PR #12, opened 2020-10-07
feat: add new model for LocalizedString type

emmenko:
This would be our first published model package. It aims to be a reference for how to build other models.


models/localized-string/src/types.ts
```diff
@@ -0,0 +1,17 @@
+import type { TBuilder } from '@commercetools-test-data/core';
+
+export type TLocalizedStringBuilder = TBuilder<TLocalizedString>;
+
+export type TCreateLocalizedStringBuilder = (args?: {
+  defaults?: Partial<TLocalizedString>;
+}) => TLocalizedStringBuilder;
+
+export type TLocalizedString = {
+  [locale: string]: string | undefined;
```
emmenko, line 10: Because of the "empty" preset, where we set the value to `undefined`, we also need to define it here as a type.

Q: is this something we have to support and is it how it's meant to be?

README.md
```diff
@@ -6,9 +6,25 @@
...
-_Coming soon_
+Each model is published as a standalone NPM package. Models are defined in the `models/*` workspace.
+
+All models are built using the `core` package, which contains the necessary methods to implement and work with models. The `core` package is very much agnostic of the commercetools platform models, so you can potentially use it to build your own models for other systems.
+
+To know more about how to work and build data models, [check out the documentation of the `core` package](./core).
+
+# Usage
+
+Using models is pretty straightforward. You import the package and build the model, which returns the data in the request shape.
```
ahmehri, line 21:
```suggestion
Using models is pretty straightforward. You import the package and build the model, which returns the data in the requested shape.
```

core/README.md
```diff
@@ -0,0 +1,160 @@
...
+
+## Install
+
+```bash
+$ yarn add -D @commercetools-test-data/core
+```
+
+## Creating a new data model
+
+Models are treated packages. Each model will be published as a standalone NPM package.
```
ahmehri, line 13: What's the rationale behind defining the models as standalone NPM packages?

core/README.md
```diff
@@ -0,0 +1,160 @@
...
+
+```ts
+const generator = Generator<TAuthor>({
+  name: 'Author',
+  fields: {
+    firstName: fake((f) => f.name.firstName()),
+    locale: 'en',
+  },
+});
+```
```
ahmehri, line 57:
```suggestion
```ts
import { fake } from '@jackfranklin/test-data-bot';

const generator = Generator<TAuthor>({
  name: 'Author',
  fields: {
    firstName: fake((f) => f.name.firstName()),
    locale: 'en',
  },
});
```
```

File: core/README.md [hunk 5]
```diff
@@ -0,0 +1,160 @@
...
+
+Builders are essentially composed by a generator and (optionally) transformers.
+
+### Generator
+
+A generator is what describes the initial model shape. Most of the fields can and should be initialized with some values. You can define random values using `fake` or static values.
+
+```ts
+const generator = Generator<TAuthor>({
+  name: 'Author',
```
ahmehri, line 51: FYI: we can [not use](https://github.com/jackfranklin/test-data-bot#creating-your-first-builder) `name`.

> Feel free to use the name property if you like, but it's not used for anything in test-data-bot. It will probably get removed in a future major version.

core/README.md
```diff
@@ -0,0 +1,160 @@
...
+  fields: {
+    firstName: fake((f) => f.name.firstName()),
+    locale: 'en',
+  },
+});
+```
+
+> The shape of the generator model is what usually is used in the UI and does not need transformation.
```
ahmehri, line 59: Not necessarily, we can use the REST or GraphQL transformation in the UI, I propose to rephrase it to

```suggestion
> The shape of the generator model is what is defined as a model in [commercetools HTTP API](https://docs.commercetools.com/api/).
```

core/README.md
```diff
@@ -0,0 +1,160 @@
...
+```ts
+const transformers = {
+  default: Transformer<TAuthor, TAuthor>('default', {
+    buildFields: ['books'],
+  }),
+  graphql: Transformer<TAuthor, TAuthorGraphql>('graphql', {
+    buildFields: ['books'],
+    addFields: () => ({
+      __typename: 'Author',
+    }),
```
ahmehri, line 108: Irrelevant.

```suggestion
```

core/README.md
```diff
@@ -0,0 +1,160 @@
...
+
+Defining a `graphql` and/or a `rest` transformer is optional and depends on what the specific data requirements are.
+
+For example, for the `Author` we can define a `graphql` transformer like this:
+
+```ts
+const transformers = {
+  default: Transformer<TAuthor, TAuthor>('default', {
+    buildFields: ['books'],
+  }),
```
ahmehri, line 131: Irrelevant.

```suggestion
```

core/README.md
```diff
@@ -0,0 +1,160 @@
...
+- `addFields`: a function that should return an object with new fields that needs to be added to the model. The object is then merged with the initial data.
+  The function gets an object argument with the initial `fields` object, in case it's needed to construct the new object.
+- `replaceFields`: a function that should return the fully transformed value. This is useful for example when transforming data into different data types, for example an object into an array of objects.
+  The function gets an object argument with the initial `fields` object, in case it's needed to construct the value.
+
+> The `replaceFields` option is mutually exclusive in regards to the `addFields` and `removeFields` options.
+
+### Presets
+
+Presets are pre-configured models (builders) that can be defined for specific use cases.
```
ahmehri, line 156: This may be outside the scope of the PR but it's worth mentioning. It's important to document how [presets should be exported as functions to ensure the randomness of data](https://github.com/commercetools/merchant-center-frontend/tree/master/packages-shared/test-data#presets-1).

core/README.md
```diff
@@ -0,0 +1,160 @@
...
+
+To create a new model, add a new package in the `models` workspace with the following structure:
+
+```
+models/<package_name>
+├── LICENSE
+├── README.md
+├── index.ts
+├── package.json
+├── src
```
ahmehri, line 23: What about `constants.ts` file that contains the constants?

models/localized-string/package.json
```diff
@@ -0,0 +1,43 @@
...
+  "version": "0.1.0",
+  "description": "Data model for LocalizesString type",
+  "bugs": "https://github.com/commercetools/test-data/issues",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/commercetools/test-data.git",
+    "directory": "models/localized-string"
+  },
+  "keywords": [
+    "javascript",
```
ahmehri, line 12:
```suggestion
    "typescript",
```

models/localized-string/package.json
```diff
@@ -0,0 +1,43 @@
...
+  ],
+  "license": "MIT",
+  "private": false,
+  "publishConfig": {
+    "access": "public"
+  },
+  "main": "./dist/localized-string-index.cjs.js",
+  "module": "./dist/localized-string-index.es.js",
+  "typings": "./dist/typings/index.d.ts",
+  "types": "./dist/typings/index.d.ts",
```
ahmehri, line 23: Just wondering what's the difference between `typings` and `types`?

models/localized-string/src/generator.ts
```diff
@@ -0,0 +1,18 @@
+import type { TLocalizedString } from './types';
+
+import { Generator, fake } from '@commercetools-test-data/core';
+
+/**
+ * A localized string is a JSON object where the keys are of ↗ IETF language tag,
+ * and the values the corresponding strings used for that language.
+ */
```
ahmehri, line 8: Wondering if the convention should be a JSDoc comment or just a comment pointing to the docs. We have both variants in mc test-data.

```suggestion
// https://docs.commercetools.com/api/types#localizedstring
```

models/localized-string/src/generator.ts
```diff
@@ -0,0 +1,18 @@
...
+const generator = Generator<TLocalizedString>({
+  name: 'LocalizedString',
+  fields: {
+    de: fake((f) => f.lorem.words(5)),
+    en: fake((f) => f.lorem.words(3)),
```
ahmehri, line 13: Shouldn't `en` be the first?

models/localized-string/package.json
```diff
@@ -0,0 +1,43 @@
...
+    "build": "yarn build:bundles && yarn build:typings",
+    "build:bundles": "cross-env NODE_ENV=production rollup -c ../../rollup.config.js -i ./src/index.ts --dir dist",
+    "build:bundles:watch": "yarn build:bundles -w",
+    "build:typings": "cross-env tsc -p tsconfig.declarations.json --emitDeclarationOnly --declarationDir dist/typings"
+  },
+  "dependencies": {
+    "@babel/runtime": "7.11.2",
+    "@babel/runtime-corejs3": "7.11.2",
+    "@commercetools-test-data/core": "1.0.1",
+    "faker": "^4.1.0"
```
ahmehri, line 41: Shouldn't be fixed?

```suggestion
    "faker": "4.1.0"
```

README.md
```diff
@@ -6,9 +6,25 @@
...
-> At some point we will also open source the common models for the commercetools platform, ready to be used.
+This repository contains data models within the commercetools platform ecosystem.
 
-# Getting started
+> In time, we will continue to implement and open source test data models. If you are interested or missing a specific data model, feel free to open a feature request or try to contribute to the repository.
 
-_Coming soon_
+Each model is published as a standalone NPM package. Models are defined in the `models/*` workspace.
+
+All models are built using the `core` package, which contains the necessary methods to implement and work with models. The `core` package is very much agnostic of the commercetools platform models, so you can potentially use it to build your own models for other systems.
```
tdeekens, line 15:
```suggestion
All models are built using the `core` package, which contains the necessary methods to implement and work with models. The `core` package is very much agnostic of the commercetools platform domain, so you can potentially use it to build your own models for other domains.
```

core/README.md
```diff
@@ -0,0 +1,160 @@
...
+│   └── types.ts
+├── tsconfig.declarations.json
+└── tsconfig.json
+```
+
+To define and build new models, you need to use the core package `@commercetools-test-data/core`.
+
+### Builder
+
+A builder is the result of the data model. Builders allow composition with other models and to transform the data into different shapes, for example `rest` or `graphql`.
```
tdeekens, line 39:
```suggestion
A builder is the result of the data model. Builders allow compositions with other models and to transform the data into different shapes, for example `rest` or `graphql`.
```

core/README.md
```diff
@@ -0,0 +1,160 @@
...
+
+When you want to get the generated data, you would **build** the model (builder).
+
+Builders are essentially composed by a generator and (optionally) transformers.
+
+### Generator
+
+A generator is what describes the initial model shape. Most of the fields can and should be initialized with some values. You can define random values using `fake` or static values.
```
tdeekens, line 47:
```suggestion
A generator is what describes the initial model shape and data. Most of the fields can and should be initialized with some values. You can define random values using `fake` or static values.
```

models/localized-string/package.json
tdeekens: Somewhat a convention I think. At least for us. Types are external. Typings are what we build and expose as types to others.

models/localized-string/src/types.ts
tdeekens: Defines on the definition of empty I assume. If empty is explicitly set to `undefined` we, I think need it.

tdeekens (commented):
Thanks for moving this forward. Looks very promising!

core/README.md
emmenko: Just so that you can pick what you want/need.

I suppose we can maybe group the models by domain or something instead.

- `commons`: for things like https://docs.commercetools.com/api/types
- `products`: for models related to the product domain.
- `accounts`: things like user, project, etc.

Would that be reasonable? cc @tdeekens

core/README.md
emmenko: The structure here is mostly for the minimal setup. If certain models require more things, they can just do that.

core/README.md
emmenko: Ah thanks, I didn't know. I guess we can just drop it then.

@tdeekens was there a particular reason for us to define it?

core/README.md
emmenko: The `fake` is exported from our core package, so we should import from that. Thanks for the suggestion.

core/README.md
emmenko: From what @tdeekens explained to me, the shape of the model should be based on what is used in the UI.

@tdeekens can you maybe clarify this?

core/README.md
emmenko: Why is it irrelevant? 🤔

core/README.md
emmenko: Again, why is it irrelevant? 🤔

core/README.md
emmenko: Ah good point, I'll try to document this as well.

core/README.md
ahmehri: Yes I noticed that lately, the idea is to document the import.

core/README.md
ahmehri: The emphasis here is on `buildFields`, adding other stuff will make the reader destructed, in my opinion.

models/localized-string/package.json
emmenko: I chose to have it like this to avoid different versions with the `faker` dependency coming from the test-data-bot package.

Maybe we can just re-export it via the core package, that would be better maybe

models/localized-string/src/generator.ts
emmenko: Good point, a link would be enough for me.

models/localized-string/src/generator.ts
emmenko: What's the difference?

core/README.md
ahmehri: The emphasis understood from the following sentence is the GraphQL transformer. Adding other stuff can make the reader distracted, in my opinion.

> For example, for the `Author` we can define a `graphql` transformer like this:

core/README.md
ahmehri: Except that `constants` was an established convention, a unified way of defining constants in test-data. Maybe it's worth using the same convention here as well, to avoid future discussions.

emmenko: Does it mean that every model should have a `constants` file?

core/README.md
```diff
@@ -36,25 +36,26 @@ To define and build new models, you need to use the core package `@commercetools
...
+import { Generator, fake } from '@commercetools-test-data/core';
+
 const generator = Generator<TAuthor>({
   name: 'Author',
   fields: {
     firstName: fake((f) => f.name.firstName()),
     locale: 'en',
   },
 });
-```
```
ahmehri: I think this deletion was a typo.

core/README.md
emmenko: I see. I'll put a `// ...` comment then to hint that other fields can be provided.

core/README.md
emmenko: Yep ok 👌

core/README.md
ahmehri: As far as I understood, yes, it behaves as a placeholder if there are no constants, just like the `presets` folder.

models/localized-string/src/generator.ts
ahmehri: It's a nitpicking, I mean commercetools is an International company and I would assume that English is the first locale. Feel free to ignore.

core/README.md
tdeekens: I like the idea but wonder it it's generally that useful however does not hurt. If we group by domain it might be helpful. You could ask the same for babel in parts I suppose.

core/README.md
tdeekens: None. Just carried forward from the first generator. We can drop it.

core/README.md
tdeekens: Yes, I think it's reasonable that we say that the UI shape is the default. The other ones are API shapes. Basically we need to pick one as then we can map into each other. I think the default shape is the best candidate.

core/README.md
```diff
@@ -0,0 +1,204 @@
...
+import * as Book from '../../book';
+import Author from './builder';
+
+const preset = (): TAuthorBuilder =>
+  Author.random().books([Book.random().title(faker.random.words())]);
+export default preset;
+
+// presets/index.ts
+export { default as authorWithOneBook } from './author-with-one-book';
+```
```
emmenko, line 200: Here's the updated presets section

models/commons/package.json
```diff
@@ -0,0 +1,44 @@
+{
+  "name": "@commercetools-test-data/commons",
```
emmenko, line 2: Renamed the package to `commons`, which can export multiple models.

tdeekens (approved):

core/README.md
```diff
@@ -46,7 +47,7 @@ Builders are essentially composed by a generator and (optionally) transformers.
 
 A generator is what describes the initial model shape and data. Most of the fields can and should be initialized with some values. You can define random values using `fake` or static values.
 
-```ts
+````ts
```
ahmehri: Any reason for the additional backtick?

```suggestion
```ts
```

models/commons/src/index.ts
```diff
@@ -0,0 +1,7 @@
+import { type } from 'os';
```
ahmehri: Does what this import do?

ahmehri (approved):

PR #20, opened 2020-10-14
fix(localized-string): filter out empty translations

pa3:
LocalizedString can't have an undefined as it's translation value, but we use it in some cases to override one created by `random()` builder (e.g. `LocalizedString.presets().empty`). That leads to warnings in tests, using such test data.

This PR fixes this by filtering out empty translations in transform step.


ahmehri (approved):

emmenko (approved):
Thanks, I was wondering about this behavior. Makes total sense to me! 👌

Issue #23, opened 2020-10-28
Remove unused `defaults` option


emmenko:
We never really used it internally and we can manage to live without it, so we should probably remove it.

PR #24, opened 2020-10-29
feat: drop `defaults` prop from builder

pa3:
As we collected some experience with using test-data, it's turned out people never use an option to provide `defaults` prop to `Builder` instance, which current implementation provides, because facilities provided by `generator` prop are just enough. So, in order to shrink the API surface a bit, we decided to drop this prop (https://github.com/commercetools/test-data/issues/23). 

Apart from mentioned above, this PR reorganizes how `Builder` invokes the provided generator internally. Namely, there is no reason to wait for `buildXXX` or setter function to be invoked and generation can be performed right away. Which simplifies code a bit.

### DIFF
--- a/core/src/builder.ts
+++ b/core/src/builder.ts
@@ -44,10 +44,8 @@ const createState = <Model extends Json>({
   };
 };
 
-function PropertyBuilder<Model extends Json>({
-  defaults,
-}: { defaults?: Partial<Model> } = {}) {
-  const state = createState<Model>({ initial: defaults });
+function PropertyBuilder<Model extends Json>(initialProps?: Partial<Model>) {
+  const state = createState<Model>({ initial: initialProps });
 
   const builder: TPropertyBuilder<Model> = new CustomProxy<
     Partial<Model>,
@@ -59,12 +57,6 @@ function PropertyBuilder<Model extends Json>({
         switch (prop) {
           case 'get': {
             return () => state.get();
-          }
-          case 'update': {
-            return (update: Partial<Model>) => {
-              state.merge(update);
-              return builder;
-            };
           }
           default: {
             return (fnOrValue: TBuilderMapStateFunction<Model> | unknown) => {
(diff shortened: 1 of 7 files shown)


pa3:
@tdeekens @emmenko conceptually that's a breaking change, which though, should not cause any issues, as dropped functionality was not used anyways. Should I bump a major version in the changeset, WDYT?

tdeekens (commented):
I wouldn't mind major releases and for simplicity follow it. Personally, I don't think we have to fear it. Especially in the early phase of a project. 

Also cause the dependency update PRs by e.g. renovate group accordingly. Which can be helpful in case we missed something.

Regarding the `defaults` prop. I think as discussed it's fine to drop it. We however need a way to compose builders. `Builder({ defaults: Builder2() }.id('foo')` might not be the slickest API but e.g. `Buidler().from(Builder2)` could be something.

core/src/builder.ts
```diff
@@ -84,39 +76,44 @@ function PropertyBuilder<Model extends Json>({
...
   transformers,
 }: TBuilderOptions<Model> = {}): TBuilder<Model> {
-  const propertyBuilder = PropertyBuilder<Model>({ defaults });
   const applyGeneratorIfExists = (): Partial<Model> => {
     if (!generator) return {};
-    return generator.generate({ defaults });
+    return generator.generate();
   };
 
+  const propertyBuilder = PropertyBuilder<Model>(applyGeneratorIfExists());
```
emmenko, line 87: So this is where we "generate things" before hand, correct?

models/commons/src/localized-string/transformers.ts
```diff
@@ -1,15 +1,19 @@
...
 const transformers = {
   graphql: Transformer<TLocalizedString, TLocalizedStringGraphql>('graphql', {
     replaceFields: ({ fields }) =>
-      Object.entries(fields).map(([locale, value]) => ({
-        locale,
-        value,
-        __typename: 'LocalizedString',
-      })),
+      Object.entries(fields)
+        .filter(([, value]) => !isNil(value))
```
emmenko, line 11: 💯

emmenko (approved):
Thanks, especially for making the builder implementation simpler!

emmenko:
> We however need a way to compose builders. `Builder({ defaults: Builder2() }.id('foo')` might not be the slickest API but e.g. `Buidler().from(Builder2)` could be something.

Should we track this in a separate issue?

tdeekens (approved):
Thanks. Code looks good. 

Let's not use `defaults` to compose builders but let's create a separate issue for it.

models/commons/src/localized-string/transformers.ts
pa3: Hm. That's actually weird. I've already merged these changes two weeks ago: https://github.com/commercetools/test-data/pull/20/files

Was that revert?

core/src/builder.ts
pa3: Yup.

models/commons/src/localized-string/transformers.ts
pa3: Nope. Master has that change already... :thinking:

pa3: Is it some issue with master/main naming?

PR #38, opened 2020-11-14
refactor: compile and bundle monorepo using preconstruct cli

emmenko:
I saw that changesets monorepo is using `@preconstruct/cli` to manage compiling and bundling monorepos, which means:
- no more rollup (precostruct does that)
- no more ts config files in each package

image


adnasa (approved):
🔥

tdeekens (approved):

PR #44, opened 2020-12-10
refactor(models): move commons to own entry-point

adnasa:
#### Summary

- attempts to split `commons` to its own entry-point
This is a breaking change, because we have to change imports on the consumer side from `'@commercetools-test-data/commons' -> '@commercetools-test-data/models/commons'`

#### Description

> This PR is open to openly discuss how we add new models into `test-data`..

I want to try to add the Product model into `test-data`, because we might need it for the `AppKit`, given the incoming product picker.

I discovered that I might need this model.
One "simpler" approach is to define a simple product model to enable `msw` adoption for the product picker `application-kit/test-data`.

Given that the effort is the same, whether I author it in App kit or here, I decided that it makes more sense to start here.

#### Proposal

My proposal is move `commons` to its entry point so that `models` can be a space to add Product, Category etc...

- 👍 : consumers can `import { Product } from '@commercetools-test-data/models'`
- 👎  : breaking change
- 👎 : Product can not import from `commons` using relative path.

Another approach I considered is (might even be better actually, but let me know what you think)

1. move `commons` to the root folder and update the `preconstruct.packages` config under `package.json` 
2. define `models` as its own package.

This means no breaking change ;)

what do you think?

---

### TLDR;

I want to add Product, either we

1. keep `commons` "scoped" under `models`?
2. have two different packages; `test-data/commons` and `test-data/models`


package.json
```diff
@@ -4,15 +4,9 @@
...
-    "core",
-    "models/*"
-  ],
+  "workspaces": ["core", "models"],
   "preconstruct": {
-    "packages": [
-      "core",
-      "models/*"
-    ]
+    "packages": ["core", "models"]
```
adnasa, line 9: if we want, `commons` can be its own package on the root folder. it makes more sense.

models/package.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "@commercetools-test-data/models",
+  "version": "2.1.0",
+  "description": "Data model for commercetools platform common types",
```
adnasa, line 4: should update this.

emmenko:
The idea was to have packages by domain:

* `commons`: basic models, like https://docs.commercetools.com/api/types
* `products`: everything related to products, categories, etc
* ...and so on

https://github.com/commercetools/test-data/pull/12#discussion_r501613764

adnasa:
ahaaaa.. I must have been away when that discussion happened.
So this means, I can introduce `products` without changing the status quo.

thanks for clarifying. we can close this. ;)

adnasa:
I have a small concern regarding that comment however;
I believe at some point, it will start making sense to generate types for graphql models.

if we go for this, the types might have to end up as its own package too instead of inline with the model.
e.g `TProduct` is not exported from `models/products`, but rather `types` root package.

emmenko:
> it will start making sense to generate types for graphql models

What do you mean exactly by that? Can you give a concrete example? 🤔

adnasa:
> if we go for this, the types might have to end up as its own package too instead of inline with the model.
e.g TProduct is not exported from models/products, but rather types root package.

this is actually important, because even if we structure models based on their domains, a `reference` can be expanded and we're going to need an import to specify the type

adnasa:
> What do you mean exactly by that? Can you give a concrete example? 🤔

so in this example, we typed the `LocalizedString` manually.
https://github.com/commercetools/test-data/blob/main/models/commons/src/localized-string/types.ts

I'd like to avoid authoring the `Product` all by hand 😄

emmenko:
Yeah sure, makes sense.

adnasa:
I'll set that one up then, before introducing `models/products`. I need it after all :)

PR #58, opened 2021-03-15
feat(commons): add `Reference` and `Initiator`

adnasa:
### Description

I want to eventually add `Product` and `ProductType`, both of which need Initiator and Reference respectively.

### DIFF
--- a/models/commons/src/index.ts
+++ b/models/commons/src/index.ts
@@ -2,5 +2,6 @@
 export * from './localized-string/types';
 
 // Export models
+export * as Initiator from './initiator';
 export * as LocalizedString from './localized-string';
 export * as Reference from './reference';
--- a/models/commons/src/index.ts
+++ b/models/commons/src/index.ts
@@ -3,3 +3,4 @@ export * from './localized-string/types';
 
 // Export models
 export * as LocalizedString from './localized-string';
+export * as Reference from './reference';
--- a/models/commons/src/initiator/generator.ts
+++ b/models/commons/src/initiator/generator.ts
@@ -8,8 +8,8 @@ const generator = Generator<TInitiator>({
     externalUserId: fake((f) => f.random.uuid()),
     anonymousId: fake((f) => f.random.uuid()),
     clientId: fake((f) => f.random.uuid()),
-    customerRef: Reference.random().build(),
-    userRef: Reference.random().build(),
+    customerRef: Reference.random().typeId('customer').build(),
+    userRef: Reference.random().typeId('user').build(),
   },
 });
 
--- a/models/commons/src/initiator/generator.ts
+++ b/models/commons/src/initiator/generator.ts
@@ -2,6 +2,7 @@ import type { TInitiator } from './types';
 import Reference from '../reference';
 import { Generator, fake } from '@commercetools-test-data/core';
 
+// https://docs.commercetools.com/api/types#client-logging
 const generator = Generator<TInitiator>({
   fields: {
     isPlatformClient: fake((f) => f.random.boolean()),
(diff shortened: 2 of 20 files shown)


models/commons/src/initiator/generator.ts
```diff
@@ -0,0 +1,16 @@
...
+import Reference from '../reference';
+import { Generator, fake } from '@commercetools-test-data/core';
+
+const generator = Generator<TInitiator>({
+  fields: {
+    isPlatformClient: fake((f) => f.random.boolean()),
+    externalUserId: fake((f) => f.random.uuid()),
+    anonymousId: fake((f) => f.random.uuid()),
+    clientId: fake((f) => f.random.uuid()),
+    customerRef: Reference.random().build(),
```
mohib0306, line 11: don't we need to specify the `typeId=customer` here. I mean is the `typeId` optional in references

models/commons/src/index.ts
```diff
@@ -2,4 +2,6 @@
 export * from './localized-string/types';
 
 // Export models
+export * as Initiator from './initiator';
```
emmenko, line 5: What model is this exactly? Can we use a more specific name maybe?

adnasa: Initiator is used for `lastModifiedBy` and `createdBy` fields
https://docs.commercetools.com/api/types#client-logging

models/commons/src/initiator/generator.ts
adnasa: good catch!

models/commons/src/index.ts
emmenko: Hmm ok, but I'm still confused about this name. I can't find it anywhere in the docs. Shouldn't this be called `ClientLogging` or something like that? 🤔

models/commons/src/initiator/generator.ts
adnasa: https://github.com/commercetools/test-data/pull/58/commits/df91c001e90f80c4bc44857993a8befec3197d92

models/commons/src/index.ts
adnasa: As discussed, I added a description and code comment referring to the docs

models/commons/src/reference/types.ts
```diff
@@ -0,0 +1,22 @@
...
+
+type TExpandedReferenceObject = {
+  id: string;
+  [key: string]: unknown;
+};
+
+export type TReference = {
+  typeId: string;
+  id: TExpandedReferenceObject['id'];
+  obj?: TExpandedReferenceObject;
```
emmenko, line 15: It would be better to pass a generic here. If it's too complicated, we can leave it like that, or just type it as `obj?: unknown`

emmenko (commented):
Looks good, I left a small suggestion.

models/commons/src/reference/types.ts
adnasa: could it be something like this?

```
export type TReference<T extends TExpandedReferenceObject> = {
  typeId: string;
  id: T['id'];
  obj?: T;
};
```

emmenko: Something like that. The `id` can simply be typed as a `string`

mohib0306 (approved):

models/commons/src/reference/types.ts
adnasa: https://github.com/commercetools/test-data/pull/58/commits/2986fa323bf13fb7ac0280aa3025c101cdee9d46

models/commons/src/reference/types.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export type TCreateReferenceBuilder = () => TReferenceBuilder;
+
+type TExpandedReferenceObject = {
+  id: string;
+  [key: string]: unknown;
+};
+
+export type TReference<
+  T extends TExpandedReferenceObject = TExpandedReferenceObject
```
tdeekens, line 13: Nice!

tdeekens (approved):

PR #67, opened 2021-04-13
refactor: use faker v5

emmenko:
Given that `@jackfranklin/test-data-bot` seems unmaintained and still uses `faker` v4, I decided to drop its dependency and include the source code directly here, so that we can use `faker` v5 at our own pace.

We also had the workaround with the types (https://github.com/jackfranklin/test-data-bot/pull/344).

So essentially, we can now ship `faker` v5.


tdeekens (approved):
Thanks. I was thinking the same thing the other day.

adnasa (approved):
yeeeeeah boooiiii

PR #72, opened 2021-04-20
Add `models/category`

adnasa:
### Summary

- introduces new package `@commercetools-test-data/utils` to export some helpers
- adds `models/category`

### DIFF
--- a/models/category/LICENSE
+++ b/models/category/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2020 commercetools GmbH
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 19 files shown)


models/category/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2020 commercetools GmbH
```
adnasa, line 3: yikes.. will update

models/category/src/presets/index.ts
```diff
@@ -0,0 +1,3 @@
+export default {
+  // @TODO: add `category-draft`
```
adnasa, line 2: although it is nothing I would do yet,
it makes a lot more sense to add the draft models in the same package as the main model instead of having separate packages.

what do you think?
cc @ahmehri @tdeekens

ahmehri: No hard opinion here, I mean I get it, it makes sense to group both the main and draft models together, which is better from a code colocation point of view. I don't see a reason why to not do it.

adnasa:
With a simple model, I consider this ready for review.

utils/src/create-related-dates.ts
```diff
@@ -0,0 +1,47 @@
...
+ * @param {Number} [recentDaysFromPastReference=10] - the older date will be in
+ * some time from the reference to recentDaysFromPastReference days ago.
+ * @returns {Array} callbacks
+ * {Function} callbacks[0] - the function that will retrieve the older date,
+ * it's expecting to be invoked with faker as parameter
+ * {Function} callbacks[1] - the function that will retrieve the newer date,
+ * it's expecting to be invoked with faker as parameter
+ */
+const createRelatedDates = (recentDaysFromPastReference: number = 10) => {
+  const pastReference = Date.now();
```
tdeekens, line 34: I recently changed this to this

```suggestion
  const pastReference = new Date().toISOString();
```

to avoid getting invalid dates. `Date.now()` returns in int while faker uses `Date.parse` internally. Date.parse as per MDN expects an ISO date string.

tdeekens (approved):
Thanks! I like that we're exporting the createRelatedDates then we can share the same internally.

Issue #85, opened 2021-07-06
`buildGraphqlList` generates a wrong `__typename`


ahmehri:
**Describe the issue**

`buildGraphqlList` generates a wrong `__typename`.

**To Reproduce**

Use `buildGraphqlList` to generate a GraphQL paginated result of e.g `CategorySearch`:

```js
buildGraphqlList([CategorySearch.random()], {
  total: 0,
  name: 'CategorySearch',
})
```

**Current behavior**

The `__typename` of the paginated result follows the naming convention `<name>QueryResult`, e.g `CategorySearchQueryResult`.

**Expected behavior**

The `__typename` of the paginated result should follow the naming convention `<name>Result`, e.g `CategorySearchResult`.

**Additional context**

This doesn't cause any problems (other than inconsistency) when having GraphQL queries without fragments. However, the `__typename` is important when using fragments. Not generating the correct one will cause problems such as not returning mock data when using msw.

https://github.com/commercetools/test-data/blob/4799b64763613613e7c9dc2cf74401fe4d31a8b8/core/src/helpers.ts#L112-L120

ahmehri:
I just found out that for other queries, the `<name>QueryResult` naming convention is used, e.g for the `cartDiscounts` query it's `CartDiscountQueryResult`. This means that we have a mix of used naming conventions forced by the GraphQL API (naming inconsistency).

The only way that I can see to solve this is to let consumers provide the correct `__typename` instead of generating it automatically.

PR #87, opened 2021-07-06
feat(core): to add typename support to graphql list building

tdeekens:
- Closes https://github.com/commercetools/test-data/issues/85

The exact `__typename` on any list within all services and parts of GraphQL schemas of the CT can't be inferred by `${name}QueryResult`. 

As a solution, I propose to have an explicit `typename` overwrite. Over time we could even argue to remove the `name` altogether.

Closes https://github.com/commercetools/test-data/issues/85.

### DIFF
--- a/core/src/types.ts
+++ b/core/src/types.ts
@@ -26,7 +26,7 @@ export type TPaginatedQueryResult<Model extends Json> = {
 
 export type TGraphqlPaginatedQueryResultOptions = {
   name?: string;
-  typename?: string;
+  __typename?: string;
 } & TPaginatedQueryResultOptions;
 
 export type TGraphqlPaginatedQueryResult<Model extends Json> = {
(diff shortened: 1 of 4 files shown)


core/src/types.ts
```diff
@@ -25,7 +25,8 @@ export type TPaginatedQueryResult<Model extends Json> = {
 };
 
 export type TGraphqlPaginatedQueryResultOptions = {
-  name: string;
+  name?: string;
```
tdeekens, line 28: One could type this slightly better with a union and intersection type but to be honest I felt this is good enough.

ahmehri (approved):
Thanks for doing this! maybe one slight nit suggestion is to use `__typename` everywhere. This will relieve the developer from typing `__typename` when `typename` is expected instead.

tdeekens:
@ahmehri you mean to call the property `__typename`?

ahmehri:
Yes, exactly. Thanks ❤️ .

PR #113, opened 2022-03-27
feat: add models for `Channel`, `ChannelDraft`, `Address`

emmenko:
Ref: https://github.com/commercetools/fe-chapter-notes/issues/203

In addition to that, I refined a bit the other models and types as there were some inconsistencies and missing things.

### DIFF
--- a/models/channel/LICENSE
+++ b/models/channel/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2021 commercetools GmbH
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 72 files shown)


tdeekens (approved):

models/channel/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2021 commercetools GmbH
```
mohib0306, line 3: 2022?

mohib0306 (approved):

models/channel/LICENSE
emmenko: Yeah I wanted to do this as a follow up. Thanks for the reminder.

PR #119, opened 2022-04-22
chore: upgrade dependencies

emmenko:


models/commons/src/address/generator.ts
```diff
@@ -5,7 +5,7 @@ const generator = Generator<TAddress>({
   fields: {
     id: fake((f) => f.random.alphaNumeric(8)),
     key: fake((f) => f.lorem.slug(2)),
-    title: fake((f) => f.name.title()),
+    title: fake((f) => f.name.jobTitle()),
```
emmenko, line 8:
```
Error: [@faker-js/faker]: faker.name.title() is deprecated since v6.1.2 and will be removed in v7.0.0. Please use faker.name.jobTitle() instead.
```

ahmehri (approved):
I wonder if Renovate is configured in this repo.

emmenko:
Yes it is

PR #127, opened 2022-05-09
feat(commons): add address draft model, improve readme

emmenko:
Still part of the "fixed" mode evaluation.

Just to add some changes to the `commons` package, I thought to add a new `AddressDraft` model.


tdeekens (commented):
Internally we had this discussion having two options for drafts:

1. Having a draft preset
2. Having a draft model

We decided to use a draft model back then. Also if it entails duplication. Generally cause at times draft presets got a bit complex if I recall correctly. I think we should say on one standard way of going about this WDYT?

emmenko:
Sure, you mean like here? https://github.com/commercetools/test-data/tree/main/models/channel/src/channel-draft

tdeekens:
> Sure, you mean like here? [https://github.com/commercetools/test-data/tree/main/models/channel/src/channel-draft](https://github.com/commercetools/test-data/tree/main/models/channel/src/channel-draft?rgh-link-date=2022-05-09T11%3A11%3A51Z)

Yip. I just vaguely remember the team having a discussion around it prior. I DM'ed you with a internal piece of documentation.

tdeekens (approved):
Thanks 💯

PR #147, opened 2022-08-09
Faker: replacing deprecated fields

NickDevG:
# Faker: replacing deprecated field

Screenshot 2022-08-09 at 22 52 31

Screenshot 2022-08-09 at 22 49 03
s


emmenko (approved):
Thanks!

PR #148, opened 2022-08-10
chore: fail tests on console warn/error

emmenko:
This should help catching these deprecation warnings earlier.


NickDevG (approved):
I was thinking about this also. Thanks for adding it

PR #150, opened 2022-08-10
Localized String: return null instead of undefined for empty values

NickDevG:
# Localized String: return null instead of undefined for empty values

Our GraphQL queries return `null` when an entity doesn't have a localized value

Screenshot 2022-08-10 at 12 13 41

For consistency we should also return `null` instead of `undefined`

### DIFF
--- a/models/category/src/types.ts
+++ b/models/category/src/types.ts
@@ -22,6 +22,6 @@ export type TCategoryGraphql = Omit<
   __typename: 'Category';
   createdBy: TClientLoggingGraphql;
   lastModifiedBy: TClientLoggingGraphql;
-  nameAllLocales: TLocalizedStringGraphql | null;
-  descriptionAllLocales: TLocalizedStringGraphql | null;
+  nameAllLocales?: TLocalizedStringGraphql | null;
+  descriptionAllLocales?: TLocalizedStringGraphql | null;
 };
--- a/models/category/src/types.ts
+++ b/models/category/src/types.ts
@@ -22,6 +22,6 @@ export type TCategoryGraphql = Omit<
   __typename: 'Category';
   createdBy: TClientLoggingGraphql;
   lastModifiedBy: TClientLoggingGraphql;
-  nameAllLocales?: TLocalizedStringGraphql;
-  descriptionAllLocales?: TLocalizedStringGraphql;
+  nameAllLocales: TLocalizedStringGraphql | null;
+  descriptionAllLocales: TLocalizedStringGraphql | null;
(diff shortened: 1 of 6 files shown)


models/category/src/types.ts
```diff
@@ -22,6 +22,6 @@ export type TCategoryGraphql = Omit<
   __typename: 'Category';
   createdBy: TClientLoggingGraphql;
   lastModifiedBy: TClientLoggingGraphql;
-  nameAllLocales?: TLocalizedStringGraphql;
-  descriptionAllLocales?: TLocalizedStringGraphql;
+  nameAllLocales: TLocalizedStringGraphql | null;
+  descriptionAllLocales: TLocalizedStringGraphql | null;
```
emmenko, line 26: Looking at the generated types, optional fields should still get the `?` (and additionally the `null`)

https://github.com/commercetools/merchant-center-application-kit/blob/7dabb2768017de78a47a9a6c4673922834aabe12/packages/application-shell/src/types/generated/ctp.ts#L2073

```suggestion
  nameAllLocales?: TLocalizedStringGraphql | null;
  descriptionAllLocales?: TLocalizedStringGraphql | null;
```

models/commons/src/localized-string/helpers.ts
```diff
@@ -3,7 +3,7 @@ import { TLocalizedStringGraphql } from './types';
 
 const toLocalizedField = <Model>(value?: Model) => {
   if (!value) {
-    return undefined;
+    return null;
```
emmenko, line 6: Makes sense, thanks!

models/channel/src/types.ts
```diff
@@ -31,8 +31,8 @@ export type TChannelGraphql = Omit<
   __typename: 'Channel';
   createdBy: TClientLoggingGraphql;
   lastModifiedBy: TClientLoggingGraphql;
-  nameAllLocales?: TLocalizedStringGraphql;
-  descriptionAllLocales?: TLocalizedStringGraphql;
+  nameAllLocales: TLocalizedStringGraphql | null;
+  descriptionAllLocales: TLocalizedStringGraphql | null;
```
emmenko, line 35:
```suggestion
  nameAllLocales?: TLocalizedStringGraphql | null;
  descriptionAllLocales?: TLocalizedStringGraphql | null;
```

emmenko (commented):
Thanks (again). Returning `null` makes sense yes

NickDevG:
@emmenko any other change that we need to do?

PR #155, opened 2022-09-06
feat(standalone-prices): add standalone prices model

fuchodeveloper:


fuchodeveloper:
@dogayuksel this is the PR for adding standalone price test data. It's still in `Draft` but I'm pinging you here if you want to stay informed of the progress.

PR #159, opened 2022-10-12
⚡️ chore: migrate to `pnpm` 🚀 

emmenko:


tdeekens (commented):
What differences do you see? I tried to use nmMode in yarn and did some benchmarking and it didn't influence speed. I never had an issue with yarn3 on personal and work projects. Is it mainly switching the dependency manager to what everybody likes these days? What I'd like us do evaluate is if we want to use pnpm everywhere which could be beneficial. Cause now it's quite easy to communicate that the yarn setup is the same across almost all repos. Personally, I like changing a dependency manager where we see a benefit to do so (there might be). I don't see it being a benefit for "playing" and curiosity sake and cause of wider community "hype". I don't wanna be the party pooper but I think we should then follow through with all repos at least. Is that you goal?

.github/workflows/release.yml
```diff
@@ -24,19 +27,25 @@ jobs:
...
-      - name: Get yarn cache
-        id: yarn-cache
-        run: echo "::set-output name=dir::$(yarn config get cacheFolder)"
+      - name: Install pnpm
+        uses: pnpm/action-setup@v2.2.3
+        id: pnpm-install
+        with:
+          run_install: false
+
+      - name: Get pnpm store directory
```
tdeekens, line 36: I think the cache-action might support this natively. https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#caching-packages-data

emmenko: Thanks, I'll try that

tdeekens (commented):
Looks good. I think we should try to integrate internally in MC FE or services repo. Then we have large and small repos in scope of an investigation.

tdeekens (approved):
Sure. Why not :) We have a plan now.

emmenko:
ℹ️ Will merge next week.

PR #163, opened 2022-11-07
PRC-2500: Add standalone prices and associated entities

dogayuksel:


dogayuksel:
For faster iterations during development, will use/develop the standalone price modal locally in the prices repo. Once it is mature enough, will it migrate here.

PR #169, opened 2022-12-27
fix(category): add slug preset

valoriecarli:
The following changes have been made:

`orderHint` needs to be a string, not a number.
--> this has been updated for both Category & CategoryDraft generators and included in both build specs.

slug needs to reference `LocalizedString.presets.ofSlugs()`, not just `LocalizedString`
--> this has been updated for both Category & CategoryDraft generators


ByronDWall (approved):
changes look good to me, dont forget to add a chageset then I think this will be good to go

PR #175, opened 2023-01-11
Add Zone(w/ ZoneDraft) model

valoriecarli:
Adding Zone/ZoneDraft data & corresponding specs

### DIFF
--- a/.changeset/three-rats-shake.md
+++ b/.changeset/three-rats-shake.md
@@ -1,0 +1,6 @@
+---
+'@commercetools-test-data/channel': minor
+'@commercetools-test-data/zone': minor
+---
+
+add zone model


models/zone/package.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "@commercetools-test-data/zone",
+  "version": "4.2.2",
```
tdeekens, line 3: Isn't this 0.0.0 for now?

tdeekens (commented):
Looks good mainly wondering if we should set the pkg.json to 0.0.0 for zones and add a major release for it so it becomes 1.0.0.

.changeset/three-rats-shake.md
```diff
@@ -0,0 +1,6 @@
+---
+'@commercetools-test-data/channel': minor
+'@commercetools-test-data/zone': minor
```
tdeekens, line 3:
```suggestion
'@commercetools-test-data/zone': major
```

emmenko: I think this will bump all to major, as we are using `fixed` mode in Changeset.

As it is now it should be fine. The package will be published as version `4.x`.

tdeekens (approved):

PR #179, opened 2023-01-31
Fct 210 sample data models attribute plain enum type

tylermorrisford:
Adds the `attribute-plain-enum-value` model and spec.


.changeset/red-ducks-battle.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-test-data/plain-enum-value': minor
```
jaikamat, line 2: Does this changeset have to be regenerated now that the directory name has changed? Honest question, I'm not sure :)

tdeekens: Package name matters here. What has changed?

tdeekens: It has to change as pkg name changed.

tylermorrisford: Yes, slight whiff there on my part.

.changeset/olive-jars-tap.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-test-data/attribute-plain-enum-value': minor
+---
+
+Renamed `plain-enum-value` to `attribute-plain-enum-value`
```
stephsprinkle, line 5: I don't think we want two changesets here, better to update the previous one.

.changeset/red-ducks-battle.md
stephsprinkle, line 2:
```suggestion
'@commercetools-test-data/attribute-plain-enum-value': minor
```

.changeset/red-ducks-battle.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+create plain-enum-value model
```
stephsprinkle, line 5:
```suggestion
Add model for `AttributePlainEnumValue`.
```

stephsprinkle (approved):
Everything looks good to me, just some changeset suggestions

PR #199, opened 2023-02-14
chore(test-data): create `ProductData`

jaikamat:
## Summary

This PR creates the [ProductData](https://docs.commercetools.com/api/projects/products#productdata) model. 

## Description

The GraphQL entity differed quite substantially from the REST model. Notably, its `categories` field was of a completely different type - so I had to perform a full field replacement in the transformer since I couldn't call both `addFields` and `replaceFields` simultaneously due to type mismatches (reviewers, if you see a better way, let me know!)

### DIFF
--- a/models/product-data/README.md
+++ b/models/product-data/README.md
@@ -1,0 +1,25 @@
+# @commercetools-test-data/product-data
+
+This package provides the data model for the commercetools platform `ProductData` type
+
+https://docs.commercetools.com/api/types#centprecisionmoney
+
+# Install
+
+```bash
+$ yarn add -D @commercetools-test-data/product-data
+```
+
+# Usage
+
+```ts
+import type {
+  TProductData,
+  TProductDataDraft,
+} from '@commercetools-test-data/product-data';
+import * as ProductData from '@commercetools-test-data/product-data';
+
+const productData = ProductData.random().build<TProductData>();
+const productDataDraft =
+  ProductData.ProductDataDraft.random().build<TProductDataDraft>();
+```
--- a/models/product-data/src/generator.ts
+++ b/models/product-data/src/generator.ts
@@ -1,0 +1,25 @@
+import { LocalizedString, Reference } from '@commercetools-test-data/commons';
+import { fake, Generator } from '@commercetools-test-data/core';
+import { TProductData } from './types';
+
+// https://docs.commercetools.com/api/projects/products#productdata
+
+const generator = Generator<TProductData>({
+  fields: {
+    name: fake(() => LocalizedString.random()),
+    categories: fake(() => [Reference.presets.category()]),
+    categoryOrderHints: {},
+    description: fake(() => LocalizedString.random()),
+    slug: fake(() => LocalizedString.presets.ofSlugs()),
+    metaTitle: fake(() => LocalizedString.random()),
+    metaDescription: fake(() => LocalizedString.random()),
+    metaKeywords: fake(() => LocalizedString.random()),
+    // TODO: Include random ProductVariant when available
+    masterVariant: null,
+    variants: [],
+    // TODO: Include random SearchKeywords when available
+    searchKeywords: null,
+  },
+});
+
+export default generator;
--- a/models/product-data/src/transformers.ts
+++ b/models/product-data/src/transformers.ts
@@ -1,0 +1,102 @@
+import {
+  CategoryReference,
+  ProductVariant,
+  SearchKeyword,
+} from '@commercetools/platform-sdk';
+import * as Category from '@commercetools-test-data/category';
+import { TCategoryGraphql } from '@commercetools-test-data/category';
+import { LocalizedString, Reference } from '@commercetools-test-data/commons';
+import { Transformer } from '@commercetools-test-data/core';
+import { faker } from '@faker-js/faker';
+import type { TProductData, TProductDataGraphql } from './types';
+
+const transformers = {
+  default: Transformer<TProductData, TProductData>('default', {
+    buildFields: [
+      'name',
+      'categories',
+      'description',
+      'slug',
+      'metaTitle',
+      'metaDescription',
+      'metaKeywords',
+    ],
+  }),
+  rest: Transformer<TProductData, TProductData>('rest', {
+    buildFields: [
+      'name',
+      'categories',
+      'description',
+      'slug',
+      'metaTitle',
+      'metaDescription',
+      'metaKeywords',
+    ],
+  }),
+  graphql: Transformer<TProductData, TProductDataGraphql>('graphql', {
+    buildFields: [
+      'name',
+      'categories',
+      'description',
+      'slug',
+      'metaTitle',
+      'metaDescription',
+      'metaKeywords',
+    ],
+    /**
+     * We choose `replaceFields` because the type of the `categories` field differs
+     * between REST and GraphQL, and it must be replaced.
+     */
+    replaceFields: ({ fields }) => {
+      const nameAllLocales = LocalizedString.toLocalizedField(fields.name);
+      const descriptionAllLocales = LocalizedString.toLocalizedField(
+        fields.description
+      );
+      const slugAllLocales = LocalizedString.toLocalizedField(fields.slug);
+      const metaTitleAllLocales = LocalizedString.toLocalizedField(
+        fields.metaTitle
+      );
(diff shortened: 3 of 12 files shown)


models/product-data/src/generator.ts
```diff
@@ -0,0 +1,25 @@
+import { LocalizedString, Reference } from '@commercetools-test-data/commons';
+import { fake, Generator } from '@commercetools-test-data/core';
+import { TProductData } from './types';
+
+// https://docs.commercetools.com/api/projects/products#productdata
+
+const generator = Generator<TProductData>({
+  fields: {
+    name: fake(() => LocalizedString.random()),
+    categories: fake(() => [Reference.presets.category()]),
```
stephsprinkle, line 10: TIL the reference model is done 🙌

models/product-data/src/transformers.ts
```diff
@@ -0,0 +1,102 @@
...
+import { LocalizedString, Reference } from '@commercetools-test-data/commons';
+import { Transformer } from '@commercetools-test-data/core';
+import { faker } from '@faker-js/faker';
+import type { TProductData, TProductDataGraphql } from './types';
+
+const transformers = {
+  default: Transformer<TProductData, TProductData>('default', {
+    buildFields: [
+      'name',
+      'categories',
```
stephsprinkle, line 17: Should `categoryOrderHints`, `masterVariant` and `searchKeywords` be included here?

jaikamat: `categoryOrderHints` doesn't have a model, but I'll definitely include `masterVariant` and `searchKeywords`!

See ed791e8ff8fa66a4ec517db7b23bd5fd94fb63ea

models/product-data/src/types.ts
```diff
@@ -0,0 +1,43 @@
...
+import { TLocalizedStringGraphql } from '@commercetools-test-data/commons';
+import type { TBuilder } from '@commercetools-test-data/core';
+import { ValueOf } from '@commercetools-test-data/core/src/@jackfranklin/test-data-bot';
+
+export type TProductData = ProductData;
+
+// This type only appears in the GraphQL representation
+export type TCategoryOrderHintGraphql = {
+  categoryId: keyof CategoryOrderHints;
+  orderHint: ValueOf<CategoryOrderHints>;
```
tylermorrisford, line 18: TIL!

tylermorrisford (approved):

models/product-data/src/transformers.ts
```diff
@@ -0,0 +1,103 @@
...
+      const slugAllLocales = LocalizedString.toLocalizedField(fields.slug);
+      const metaTitleAllLocales = LocalizedString.toLocalizedField(
+        fields.metaTitle
+      );
+      const metaKeywordsAllLocales = LocalizedString.toLocalizedField(
+        fields.metaKeywords
+      );
+      const metaDescriptionAllLocales = LocalizedString.toLocalizedField(
+        fields.metaDescription
+      );
```
stephsprinkle, line 78: Can we use `buildField(s)` here too?

models/product-data/src/generator.ts
```diff
@@ -0,0 +1,38 @@
...
+    variant: null,
+    // TODO: Include random ProductVariant[] when available
+    variants: [],
+    // TODO: Include random ProductVariant[] when available
+    allVariants: [],
+    // TODO: Include random SearchKeywords when available
+    searchKeywords: null,
+    // TODO: Include random SearchKeywords[] when available
+    searchKeyword: [],
+    categoriesRef: fake(() => [Reference.presets.category()]),
```
stephsprinkle, line 32: Should this field be built in the transformed fields based on the values in `categories`? If not, they won't match. Perhaps we don't care.

models/product-data/src/generator.ts
```diff
@@ -0,0 +1,38 @@
...
+    categoryOrderHint: fake((f) => f.lorem.word()),
```
stephsprinkle, line 33: Similar question here, should this be constructed from a value in `categoryOrderHints`?

models/product-data/src/transformers.ts
```diff
@@ -0,0 +1,103 @@
...
+        categories
+      ).map((category) => ({
+        id: category.id,
+        typeId: 'category',
+      }));
+
+      /**
+       * We cannot use `replaceFields` in conjunction with `buildFields`,
+       * so we need to explicity build fields where necessary.
+       */
```
jaikamat, line 48: @emmenko Just wanted to make sure this is expected behavior

models/product-data/src/transformers.ts
```diff
@@ -0,0 +1,103 @@
...
+      return {
+        // Un-built fields with no model dependencies
+        ...fields,
+        // These have model dependencies and must be built
+        name: buildField(fields.name),
+        description: buildField(fields.description),
+        slug: buildField(fields.slug),
+        metaTitle: buildField(fields.metaTitle),
+        metaDescription: buildField(fields.metaDescription),
+        metaKeywords: buildField(fields.metaKeywords),
```
stephsprinkle, line 58: Do we need the `rest` parameter here?

models/product-data/README.md
```diff
@@ -0,0 +1,20 @@
...
+$ yarn add -D @commercetools-test-data/product-data
+```
+
+# Usage
+
+```ts
+import type { TProductData } from '@commercetools-test-data/product-data';
+import * as ProductData from '@commercetools-test-data/product-data';
+
+const productData = ProductData.random().build<TProductData>();
```
stephsprinkle, line 19: Given the new types, is this the correct usage?

models/product-data/src/generator.ts
jaikamat: Great thinking, decided to care 🤣 See 1e52392ab548a1ebf0f65ef079c589af9ded57cf

models/product-data/src/generator.ts
jaikamat: I like the way you think. Addressed in 252956b93c1fc7776c9dde58e0bc5a39beb2d819

models/product-data/src/transformers.ts
jaikamat: Works for me, it's more explicit. See 4cc8c8796537c1b78b0b145f80b28b61b3ac83c3

models/product-data/src/transformers.ts
jaikamat: Tried to timebox myself, but there are some weird TS errors that crop up when just running `buildFields` for a localized string graphql output. The helper `toLocalizedField` function is [casting](https://github.com/commercetools/test-data/blob/main/models/commons/src/localized-string/helpers.ts#L12) `as unknown as TLocalizedStringGraphql` internally, so we'd have to repeat that here on each line.

stephsprinkle: Care amount so low here

models/product-data/README.md
jaikamat: Adjusted in 47033449c4547cbc1ac2cf003703b877bb73e07c

stephsprinkle (approved):

models/product-data/src/transformers.ts
emmenko: In theory building fields can always be done. See implementation https://github.com/commercetools/test-data/blob/e74a09a2a55e258d7c63cf626d58c7e4978ca3c9/core/src/transformer.ts#L20-L39

emmenko: Ah I think we need to pass here the `transformedFields`

https://github.com/commercetools/test-data/blob/e74a09a2a55e258d7c63cf626d58c7e4978ca3c9/core/src/transformer.ts#L60

We can look into that separately. Thanks for bringing it up.

emmenko: Or feel free to try it out here first

jaikamat: This tracks with some tweaking I did on my branch. I was able to pass `fieldsReplacer({ fields: transformedFields })` which would give the replacer access to the built fields 👍 

I can make that change as well. Existing tests pass so it seems like it wouldn't cause regression

jaikamat: @emmenko Resolved in 01702640abe122afd53c66a6578f26f32a6243b9

emmenko (approved):
Awesome work! 🥳

tdeekens (approved):

PR #203, opened 2023-02-16
chore(test-data): create `Attribute` model

jaikamat:
## Summary

This PR creates the [Attribute](https://docs.commercetools.com/api/projects/products#attribute) model. This relates to FCT-188 as it is a dependent model for `ProductVariant`.

## Description

The patterns I used here deviate from our norm of representing REST models with the generator type, because the REST representation and graphql representation differ (graphql representation pictured below).
image

I chose to rename the generator type to `TModelDefault`, in line with our tests, and keep the REST type aliased to the SDK type as usual, so that the type API's for these npm modules remain consistent.

### DIFF
--- a/models/attribute/src/attribute-draft/builder.spec.ts
+++ b/models/attribute/src/attribute-draft/builder.spec.ts
@@ -38,4 +38,15 @@ describe('builder', () => {
       })
     )
   );
+
+  it(
+    ...createBuilderSpec<TAttributeDraft, TAttributeDraft>(
+      'rest',
+      AttributeDraft.random().value({ foo: 'bar' }),
+      expect.objectContaining({
+        name: expect.any(String),
+        value: { foo: 'bar' },
+      })
+    )
+  );
 });
--- a/models/attribute/src/attribute-draft/builder.spec.ts
+++ b/models/attribute/src/attribute-draft/builder.spec.ts
@@ -1,0 +1,41 @@
+/* eslint-disable jest/no-disabled-tests */
+/* eslint-disable jest/valid-title */
+import { createBuilderSpec } from '@commercetools-test-data/core/test-utils';
+import { TAttributeDraft, TAttributeDraftGraphql } from '../types';
+import * as AttributeDraft from '.';
+
+describe('builder', () => {
+  it(
+    ...createBuilderSpec<TAttributeDraft, TAttributeDraft>(
+      'default',
+      AttributeDraft.random(),
+      expect.objectContaining({
+        name: expect.any(String),
+        value: null,
+      })
+    )
+  );
+
+  it(
+    ...createBuilderSpec<TAttributeDraft, TAttributeDraft>(
+      'rest',
+      AttributeDraft.random(),
+      expect.objectContaining({
+        name: expect.any(String),
+        value: null,
+      })
+    )
+  );
+
+  it(
+    ...createBuilderSpec<TAttributeDraft, TAttributeDraftGraphql>(
+      'graphql',
+      AttributeDraft.random(),
+      expect.objectContaining({
+        name: expect.any(String),
+        value: null,
+        __typename: 'ProductAttributeInput',
+      })
+    )
+  );
+});
--- a/models/attribute/src/attribute-draft/generator.ts
+++ b/models/attribute/src/attribute-draft/generator.ts
@@ -1,0 +1,18 @@
+import { fake, Generator } from '@commercetools-test-data/core';
+import { TAttributeDraft } from '../types';
+
+/**
+ * This model does not have a corresponding REST entity,
+ * as `AttributeDraft` doesn't exist.
+ *
+ * It's been created specifically for GraphQL transformations
+ * when needing `ProductAttributeInput`.
+ */
+const generator = Generator<TAttributeDraft>({
+  fields: {
+    name: fake((f) => f.lorem.slug(3)),
+    value: null,
+  },
+});
+
+export default generator;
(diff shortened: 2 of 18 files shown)


models/attribute/src/types.ts
```diff
@@ -0,0 +1,23 @@
...
+export type TReferenceExpandable = {
+  id: String;
+};
+
+// This is the generator "catchall" representation
+export type TAttributeDefault = TAttribute & {
+  attributeDefinition: AttributeDefinition;
+  referencedResource: TReferenceExpandable;
+  referencedResourceSet: Array<TReferenceExpandable>;
+};
```
jaikamat, line 16: The generator is instead a combination of all the possible REST and graphql fields (minus `__typename`), and in the transformers, we can pick from among these

tylermorrisford: 💯

models/attribute/src/transformers.ts
```diff
@@ -0,0 +1,23 @@
...
+    buildFields: [],
+    replaceFields: ({ fields }) => ({
+      name: fields.name,
+      value: fields.value,
+    }),
+  }),
+  graphql: Transformer<TAttributeDefault, TAttributeGraphql>('graphql', {
+    buildFields: ['attributeDefinition'],
+    addFields: () => ({
+      __typename: 'RawProductAttribute',
```
stephsprinkle, line 18: So this is definitively the type used by `ProductVariant` in GraphQL, but the draft version `ProductVariantInput` uses `ProductAttributeInput` and the fields are different. Really not sure what to do here.

jaikamat: Hmm...interesting.

I'm unsure if RawProductAttribute and ProductAttributeInput have a model/draft relationship. If they do, it makes sense to pull ProductAttributeInput in as sort of a faux-draft, but again not sure.

Worst case scenario is to create ProductAttributeInput separately, and refactor as needed once we get in some more questions with Nicola 🤷‍♂️ Does that seem appropriate?

stephsprinkle: I feel like `ProductAttributeInput` is more similar to the REST `Attribute` model and `RawProductAttribute` is a weird GraphQL thing

stephsprinkle: But also, `RawProductAttribute` is the model and `ProductAttributeInput` is the draft in GraphQL land

jaikamat: image

So I suppose we do need a draft here, but _only_ for the graphql side of things, because `ProductAttributeInput` is (for graphql land) a draft.

Something I'll touch base with Nicola on, we haven't really encountered this before iirc!

jaikamat: image

Resolved: we will add a "dummy" `AttributeDraft` that will generate the appropriate graphql transformation to `ProductVariantInputAttributes`. This is for consistency with existing test-data repo structure.

jaikamat: @stephsprinkle @tylermorrisford [This commit](https://github.com/commercetools/test-data/pull/203/commits/1a8a96a3650b772b53004d14b4ddf471b1161573) addresses these findings.

models/attribute/src/attribute-draft/generator.ts
```diff
@@ -0,0 +1,18 @@
...
+ * This model does not have a corresponding REST entity,
+ * as `AttributeDraft` doesn't exist.
+ *
+ * It's been created specifically for GraphQL transformations
+ * when needing `ProductAttributeInput`.
+ */
+const generator = Generator<TAttributeDraft>({
+  fields: {
+    name: fake((f) => f.lorem.slug(3)),
+    value: null,
```
jaikamat, line 14: The documented type for this field in Impex is `String`, but in [this Slack thread](https://commercetools.slack.com/archives/C010MT0L58D/p1677163802757049), Yann mentions that it's in fact effectively the same as the HTTP `Attribute` representation (notably, it's basically `any` or `Json`)

models/attribute/src/generator.ts
```diff
@@ -0,0 +1,18 @@
...
+// https://docs.commercetools.com/api/projects/products#attribute
+
+const generator = Generator<TAttribute>({
+  fields: {
+    name: fake((f) => f.lorem.slug(3)),
+    // `value` is an overloaded field. It might be best to create your own
+    value: null,
+    attributeDefinition: fake(() => AttributeDefinition.random()),
+    referencedResource: fake((f) => ({ id: f.datatype.uuid() })),
+    referencedResourceSet: fake((f) => [{ id: f.datatype.uuid() }]),
```
emmenko, line 14: Nit: should we keep them empty / null as a default?

models/attribute/src/attribute-draft/builder.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+import * as AttributeDraft from '.';
+
+describe('builder', () => {
+  it(
+    ...createBuilderSpec<TAttributeDraft, TAttributeDraft>(
+      'default',
+      AttributeDraft.random(),
+      expect.objectContaining({
+        name: expect.any(String),
+        value: null,
```
emmenko, line 14: Should we have a test where we define a value for it?

emmenko (commented):
Nice, thanks for following it up with TeamQL.

models/attribute/src/attribute-draft/builder.spec.ts
jaikamat: Oh, good call! I never thought of writing a test outside of the boilerplate that we currently use 🧠 💡

stephsprinkle (approved):
Looks good, I like @emmenko's suggestions

tylermorrisford (approved):
LGTM!

PR #205, opened 2023-02-21
chore(test-data): create `Channels` presets

jaikamat:
## Summary

This PR creates test data presets for Channels, based on First Contact's requirements surrounding sample project data.

## Description

These models were simple and straightforward enough to configure. I opted for snapshot testing, as these are expected to be generated deterministically.

Additional comments inline.

### DIFF
--- a/models/channel/src/channel-draft/presets/hub.spec.ts
+++ b/models/channel/src/channel-draft/presets/hub.spec.ts
@@ -1,0 +1,63 @@
+import { TChannelDraft, TChannelDraftGraphql } from '../../types';
+import hub from './hub';
+
+describe('hub channel', () => {
+  it('should match REST snapshot', () => {
+    const channel = hub().build<TChannelDraft>();
+
+    expect(channel).toMatchInlineSnapshot(`
+      {
+        "address": undefined,
+        "custom": undefined,
+        "description": {
+          "de": undefined,
+          "en": "Hub",
+          "fr": undefined,
+        },
+        "geoLocation": undefined,
+        "key": "hub",
+        "name": {
+          "de": undefined,
+          "en": "Hub",
+          "fr": undefined,
+        },
+        "roles": [
+          "ProductDistribution",
+          "InventorySupply",
+        ],
+      }
+    `);
+  });
+
+  it('should match graphql snapshot', () => {
+    const channel = hub().buildGraphql<TChannelDraftGraphql>();
+
+    expect(channel).toMatchInlineSnapshot(`
+      {
+        "__typename": "ChannelDraft",
+        "address": undefined,
+        "custom": undefined,
+        "description": [
+          {
+            "__typename": "LocalizedString",
+            "locale": "en",
+            "value": "Hub",
+          },
+        ],
+        "geoLocation": undefined,
+        "key": "hub",
+        "name": [
+          {
+            "__typename": "LocalizedString",
+            "locale": "en",
+            "value": "Hub",
+          },
+        ],
+        "roles": [
+          "ProductDistribution",
+          "InventorySupply",
+        ],
+      }
+    `);
+  });
+});
--- a/models/channel/src/channel-draft/presets/hub.spec.ts
+++ b/models/channel/src/channel-draft/presets/hub.spec.ts
@@ -11,14 +11,16 @@ describe('hub channel', () => {
         "custom": undefined,
         "description": {
           "de": undefined,
-          "en": "Hub",
+          "en": undefined,
+          "en-US": "Hub",
           "fr": undefined,
         },
         "geoLocation": undefined,
         "key": "hub",
         "name": {
           "de": undefined,
-          "en": "Hub",
+          "en": undefined,
+          "en-US": "Hub",
           "fr": undefined,
         },
         "roles": [
@@ -40,7 +42,7 @@ describe('hub channel', () => {
         "description": [
           {
             "__typename": "LocalizedString",
-            "locale": "en",
+            "locale": "en-US",
             "value": "Hub",
           },
         ],
@@ -49,7 +51,7 @@ describe('hub channel', () => {
         "name": [
           {
             "__typename": "LocalizedString",
-            "locale": "en",
+            "locale": "en-US",
             "value": "Hub",
           },
         ],
--- a/models/channel/src/channel-draft/presets/index.ts
+++ b/models/channel/src/channel-draft/presets/index.ts
@@ -1,4 +1,4 @@
-import * as SampleDataFashion from './sample-data-fashion';
+import SampleDataFashion from './sample-data-fashion';
 import withInventorySupplyAndProductDistributionRoles from './with-inventory-supply-and-product-distribution-roles';
 import withInventorySupplyAndProductDistributionRolesNoAddress from './with-inventory-supply-and-product-distribution-roles-no-address';
 import withInventorySupplyRole from './with-inventory-supply-role';
--- a/models/channel/src/channel-draft/presets/index.ts
+++ b/models/channel/src/channel-draft/presets/index.ts
@@ -1,4 +1,4 @@
-import SampleDataFashion from './sample-data-fashion';
+import sampleDataFashion from './sample-data-fashion';
 import withInventorySupplyAndProductDistributionRoles from './with-inventory-supply-and-product-distribution-roles';
 import withInventorySupplyAndProductDistributionRolesNoAddress from './with-inventory-supply-and-product-distribution-roles-no-address';
 import withInventorySupplyRole from './with-inventory-supply-role';
@@ -6,7 +6,7 @@ import withInventorySupplyRoleNoAddress from './with-inventory-supply-role-no-ad
 import withProductDistributionRole from './with-product-distribution-role';
 
 const presets = {
-  SampleDataFashion,
+  sampleDataFashion,
   withProductDistributionRole,
   withInventorySupplyRole,
   withInventorySupplyRoleNoAddress,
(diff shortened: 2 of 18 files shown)


models/channel/src/channel-draft/presets/hub.spec.ts
```diff
@@ -0,0 +1,65 @@
+import { TChannelDraft, TChannelDraftGraphql } from '../../types';
+import hub from './hub';
```
jaikamat, line 2: How do we feel about this naming? I went off each entity's `key`

models/channel/src/channel-draft/presets/hub.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+  });
+
+  it('should match graphql snapshot', () => {
+    const channel = hub().buildGraphql<TChannelDraftGraphql>();
+
+    expect(channel).toMatchInlineSnapshot(`
+      {
+        "__typename": "ChannelDraft",
+        "address": undefined,
+        "custom": undefined,
```
jaikamat, line 41: I wonder if we'll need some sort of prefiltering to clear these `undefined` values when we make a query. Right now we sort of generate _everything_.

models/channel/src/channel-draft/presets/hub.spec.ts
jaikamat: Great strategy!

jaikamat: @tylermorrisford see 1283eb4c39d7ae7291fc293dfcf25d1c6722c49c 😄

models/channel/src/channel-draft/presets/sample-data-fashion/store-other.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+import storeOther from './store-other';
+
+describe('store-other channel', () => {
+  it('should match REST snapshot', () => {
+    const channel = storeOther().build<TChannelDraft>();
+
+    expect(channel).toMatchInlineSnapshot(`
+      {
+        "address": undefined,
+        "custom": undefined,
```
jaikamat, line 11: I wonder if we'll need some sort of prefiltering to clear these undefined fields when we make a query. Right now we sort of generate everything.

stephsprinkle: After checking out other preset tests, they don't use `toMatchInlineSnapshot`, which means you don't have to bother checking fields that aren't set.

models/channel/src/channel-draft/presets/sample-data-fashion/store-other.ts
```diff
@@ -0,0 +1,15 @@
+import { LocalizedString } from '@commercetools-test-data/commons';
+import { roles } from '../../../constants';
+import ChannelDraft from '../../builder';
+
+const storeOther = () =>
+  ChannelDraft()
+    .key('store_other')
+    .address(undefined)
+    .geoLocation(undefined)
+    .custom(undefined)
```
stephsprinkle, line 10: Not needed

models/channel/src/channel-draft/presets/sample-data-fashion/store-usa.ts
```diff
@@ -0,0 +1,15 @@
+import { LocalizedString } from '@commercetools-test-data/commons';
+import { roles } from '../../../constants';
+import ChannelDraft from '../../builder';
+
+const storeUsa = () =>
+  ChannelDraft()
+    .key('store_usa')
+    .address(undefined)
+    .geoLocation(undefined)
+    .custom(undefined)
```
stephsprinkle, line 10: Not needed

models/channel/src/channel-draft/presets/sample-data-fashion/hub.ts
```diff
@@ -0,0 +1,15 @@
+import { LocalizedString } from '@commercetools-test-data/commons';
+import { roles } from '../../../constants';
+import ChannelDraft from '../../builder';
+
+const hub = () =>
+  ChannelDraft()
+    .key('hub')
+    .address(undefined)
+    .geoLocation(undefined)
+    .custom(undefined)
```
stephsprinkle, line 10: Not needed

models/channel/src/channel-draft/presets/index.ts
```diff
@@ -1,10 +1,12 @@
+import * as SampleDataFashion from './sample-data-fashion';
 import withInventorySupplyAndProductDistributionRoles from './with-inventory-supply-and-product-distribution-roles';
 import withInventorySupplyAndProductDistributionRolesNoAddress from './with-inventory-supply-and-product-distribution-roles-no-address';
 import withInventorySupplyRole from './with-inventory-supply-role';
 import withInventorySupplyRoleNoAddress from './with-inventory-supply-role-no-address';
 import withProductDistributionRole from './with-product-distribution-role';
 
 const presets = {
```
stephsprinkle, line 8: Do we need to update the `channel/index.ts` file to include exports of the draft presets? Without it, what would the import look like?

models/channel/src/channel-draft/presets/sample-data-fashion/store-other.ts
jaikamat: I might have been thinking too much about implementation here, is the plan to use these in the resolvers like this:
```
import presets from '../presets'

const storeUsa = presets.SampleDataFashion.storeUsa;

// Ignore this obviously wrong syntax, it's just to illustrate a point
makeRequestToCoCo('url', {
    variables: {
        draft: {
            name: storeUsa.name,
            description: storeUsa.description,
            key: storeUsa.key,
            roles: storeUsa.roles,
        }
    }
})```

or like this

```
import presets from '../presets'

makeRequestToCoCo('url', {
    variables: {
        draft: { ...presets.SampleDataFashion.storeUsa() }
    }
})
```

I was thinking of the latter, hence my obsession with ensuring the presets were of a specific shape - but that might be too obsessive 😬

Comment (stephsprinkle) [reply models/channel/src/channel-draft/presets/sample-data-fashion/store-other.ts:10]: Definitely the latter! Those three undefined fields aren't needed because they aren't required fields and we aren't using them.

Comment (jaikamat) [reply models/channel/src/channel-draft/presets/sample-data-fashion/store-other.ts:10]: Ohhhh. I'm picking up what you're putting down. We just won't include these fields in the query, so they will be ignored.

File: models/channel/src/channel-draft/presets/index.ts (hunk 7 above)
Comment (jaikamat) [reply models/channel/src/channel-draft/presets/index.ts:8]: Addressed in 0850dc85538824e43a82848df7469440000c7a33

File: models/channel/src/channel-draft/presets/index.ts [hunk 8]
```diff
@@ -1,4 +1,4 @@
-import * as SampleDataFashion from './sample-data-fashion';
```
stephsprinkle: Why did you remove the `* as`?

jaikamat: When requiring in from other packages, `* as` had the following effect:

```
import {draftPresets, presets} from '@commercetools-test-data/channel';

presets.clothesStore

draftPresets.SampleDataFashion.default.storeOther // with `* as`
draftPresets.SampleDataFashion.storeOther // without `* as`
```

`channel-draft.presets.index.ts` already has a `default` export, so the `* as` syntax is superfluous. [These docs ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#namespace_import)helped me understand what was going on here, too

stephsprinkle: Ah! The zone presets need to be updated then because they use the `* as` syntax.

stephsprinkle (approved):

models/channel/src/channel-draft/presets/index.ts
jaikamat: Done! Addressed in 1b506a6e9f118d2e6614cc066b244fc44c9e1d2b

tdeekens (approved):

models/channel/src/channel-draft/presets/index.ts
```diff
@@ -1,10 +1,12 @@
+import SampleDataFashion from './sample-data-fashion';
 import withInventorySupplyAndProductDistributionRoles from './with-inventory-supply-and-product-distribution-roles';
 import withInventorySupplyAndProductDistributionRolesNoAddress from './with-inventory-supply-and-product-distribution-roles-no-address';
 import withInventorySupplyRole from './with-inventory-supply-role';
 import withInventorySupplyRoleNoAddress from './with-inventory-supply-role-no-address';
 import withProductDistributionRole from './with-product-distribution-role';
 
 const presets = {
+  SampleDataFashion,
```
emmenko, line 9: Nit: lowercase for consistency? `sampleDataFashion`

emmenko (approved):
Nice!

models/zone/src/zone-draft/presets/index.ts
```diff
@@ -1,4 +1,4 @@
-import * as SampleDataFashion from './sample-data-fashion';
+import SampleDataFashion from './sample-data-fashion';
```
tylermorrisford, line 2: 🙏

tylermorrisford (approved):

models/channel/src/channel-draft/presets/index.ts
jaikamat: [Sounds good!](https://github.com/commercetools/test-data/pull/205/commits/cd126a4c140b9ae7a53d78fb295079ec14cb73d1)

PR #207, opened 2023-02-24
fix(index): allow const access

valoriecarli:
Just a little tweak so this can be accessed via` product-type` presets


stephsprinkle (approved):

tylermorrisford (approved):
💯

emmenko (approved):

PR #212, opened 2023-02-28
chore(test-data): create `Cart`, `LineItem`, and associated drafts

jaikamat:
## Summary

While chonky, the models were fairly straightforward. `Cart` and `CartDraft` were 1:1 with its REST counterpart, while `LineItem` and `LineItemDraft` were only slightly transformed.

Comments inline.


models/commons/src/reference/generator.ts
```diff
@@ -6,6 +6,7 @@ const generator = Generator<TReference>({
   fields: {
     id: fake((f) => f.datatype.uuid()),
     typeId: null,
+    key: fake((f) => f.lorem.slug(2)),
```
jaikamat, line 9: We need to be able to configure `key` [like so](https://github.com/commercetools/test-data-research/blob/main/src/entity-drafts/carts.ts#L63)

models/line-item/src/generator.ts
```diff
@@ -0,0 +1,50 @@
...
+      f.datatype.number({
+        min: 1,
+      })
+    ),
+    addedAt: fake(addedAt),
+    state: [],
+    taxRate: fake(() => TaxRate.random()),
+    perMethodTaxRate: [],
+    supplyChannel: fake(() => Channel.random()),
+    distributionChannel: fake(() => Channel.random()),
```
stephsprinkle, line 39: These are references

models/line-item/src/line-item-draft/generator.ts
```diff
@@ -0,0 +1,34 @@
...
+const generator = Generator<TLineItemDraft>({
+  fields: {
+    productId: fake((f) => f.datatype.uuid()),
+    sku: fake((f) => f.lorem.word()),
+    quantity: fake((f) =>
+      f.datatype.number({
+        min: 1,
+      })
+    ),
+    variantId: fake((f) => f.datatype.uuid()),
```
stephsprinkle, line 21: These are typically numbers in a sequence, not UUID's

models/line-item/src/transformers.ts
```diff
@@ -0,0 +1,93 @@
...
+
+      const distributionChannelRef: TReferenceGraphql = Reference.random()
+        .id(fields.distributionChannel.id)
+        .typeId('channel')
+        .buildGraphql();
+
+      const productTypeRef: TReferenceGraphql = Reference.random()
+        .id(fields.distributionChannel.id)
+        .typeId('product-type')
+        .buildGraphql();
```
stephsprinkle, line 79: Should these be built from the corresponding reference type fields?

stephsprinkle (commented):
Chonky! A few notes 😊

models/commons/src/reference/generator.ts
jaikamat: It might - perhaps something consumers should be responsible for? We can be in charge of generating the appropriate fields, but implementation details might be on them 🤷‍♂️

stephsprinkle: I'm thinking about `.random()`, that will generate both in the current configuration.

jaikamat: We could leave `key` as `null`, that seems a good tradeoff. We'll have to configure it anyway. I can add a comment to document this as well

models/cart/src/cart-draft/generator.ts
```diff
@@ -0,0 +1,50 @@
...
+    custom: null,
+    locale: fake((f) => f.random.locale()),
+    deleteDaysAfterLastModification: null,
+    shippingRateInput: null,
+    origin: oneOf(...Object.values(origin)),
+    shippingMode: oneOf(...Object.values(shippingMode)),
+    customShipping: null,
+    shipping: [],
+    itemShippingAddresses: [],
+    discountCodes: fake((f) => [f.random.word()]),
```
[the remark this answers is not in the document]
jaikamat: I think this'll generate the whole discount code rather than just its string `code`, but you raise a good point that we should just `.build()` off it and access the code for completeless!

like `[DiscountCode.random().build().code]` or something 🤔  Not sure how invoking `build` in this will impact the transformer but I'll give it a shot!

models/line-item/src/transformers.ts
jaikamat: They basically are via setting the field ID, I opted for this because I was creating vanilla objects and manually adding a `__typename` when we have the `.buildGrahpql()` method to do just that

stephsprinkle: You're absolutely right, apparently I was loopy by the time I got here.

models/line-item/src/generator.ts
jaikamat: It's related to the same issue we were running into a few days ago, where the graphql version uses the whole `Channel` model and the REST version uses the reference, so the generator asks for the whole `Channel`, and we transform down to a reference for REST (and for transforming the corresponding `supplyChannelRef` and `distributionChannelRef` graphql fields)

stephsprinkle: Gotcha

models/cart/src/constants.ts
```diff
@@ -0,0 +1,41 @@
...
+export const origin = {
+  Customer: 'Customer',
+  Merchant: 'Merchant',
+  Quote: 'Quote',
+} as const;
+
+export const shippingMode = {
+  Single: 'Single',
+  Multiple: 'Multiple',
+};
```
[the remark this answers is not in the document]
tylermorrisford: 🦅 👁️

models/commons/src/reference/generator.ts
jaikamat: Resolved in 0cb84b0a178dabfb721bece3bfbcd9f0f73b5e98

models/cart/src/cart-draft/generator.ts
```diff
@@ -0,0 +1,50 @@
...
+    billingAddress: fake(() => AddressDraft.random()),
+    shippingMethod: fake(() => Reference.random().typeId('shipping-method')),
+    externalTaxRateForShippingMethod: null,
+    custom: null,
+    locale: fake((f) => f.random.locale()),
+    deleteDaysAfterLastModification: null,
+    shippingRateInput: null,
+    origin: oneOf(...Object.values(origin)),
+    shippingMode: oneOf(...Object.values(shippingMode)),
+    customShipping: null,
```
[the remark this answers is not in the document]
jaikamat: My mistake! Resolved in af26c13005b7ad3f9cb4d64c30883c968dc346c7

models/cart/src/constants.ts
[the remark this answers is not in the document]
jaikamat: yep! 48075ee5a1625d99864955521c1f95677e056b8f

models/cart/src/cart-draft/generator.ts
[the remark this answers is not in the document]
jaikamat: Just spent some time here. Seems like doing this "the right way" might be more effort than it's worth 😬

models/cart/src/generator.ts
```diff
@@ -0,0 +1,68 @@
...
+    shippingAddress: fake(() => Address.random()),
+    billingAddress: fake(() => Address.random()),
+    custom: null,
+    locale: fake((f) => f.random.locale()),
+    deleteDaysAfterLastModification: null,
+    shippingRateInput: null,
+    origin: oneOf(...Object.values(origin)),
+    shippingMode: oneOf(...Object.values(shippingMode)),
+    shipping: [],
+    itemShippingAddresses: [],
```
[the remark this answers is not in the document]
jaikamat: Definitely, good catch. Resolved in bdc247fcc243db057a6439cf31041d9c4c8415a9

models/line-item/src/line-item-draft/generator.ts
jaikamat: Does this mean we can use our `sequence()` helper here as we do for `version` fields, or are you referring to a sequence of numbers only, ex `123-456-789` or similar?

stephsprinkle: I think `sequence()` is completely appropriate here

jaikamat: Awesome! Addressed in 1ebd471388033c9990801d10e1258b03e197753f

models/cart/src/generator.ts
```diff
@@ -0,0 +1,68 @@
...
+    discountCodes: fake((f) => [f.random.word()]),
+    totalPrice: fake(() => CentPrecisionMoney.random()),
+    createdAt: fake(getOlderDate),
+    createdBy: fake(() => ClientLogging.random()),
+    lastModifiedAt: fake(getNewerDate),
+    cartState: oneOf(...Object.values(cartState)),
+    refusedGifts: fake(() => [Reference.random().typeId('cart-discount')]),
+  },
```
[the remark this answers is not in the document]
jaikamat: Thanks! Fixed in 3f62067b8a1b06ca5017bd7b159a800eeb216fad

stephsprinkle (approved):

tylermorrisford (approved):
This looks awesome 😎

PR #216, opened 2023-03-02
chore(test-data): create `Order` and `OrderFromCartDraft` models

jaikamat:
## Summary
This PR creates the above models. It also updates the order module's SDK package to include the latest, as some of the documented fields were not included in the types we were using.

### DIFF
--- a/models/order/package.json
+++ b/models/order/package.json
@@ -8,21 +8,30 @@
     "url": "https://github.com/commercetools/test-data.git",
     "directory": "models/order"
   },
-  "keywords": ["javascript", "typescript", "test-data"],
+  "keywords": [
+    "javascript",
+    "typescript",
+    "test-data"
+  ],
   "license": "MIT",
   "publishConfig": {
     "access": "public"
   },
   "main": "dist/commercetools-test-data-order.cjs.js",
   "module": "dist/commercetools-test-data-order.esm.js",
-  "files": ["dist", "package.json", "LICENSE", "README.md"],
+  "files": [
+    "dist",
+    "package.json",
+    "LICENSE",
+    "README.md"
+  ],
   "dependencies": {
     "@babel/runtime": "^7.17.9",
     "@babel/runtime-corejs3": "^7.17.9",
     "@commercetools-test-data/commons": "4.4.0",
     "@commercetools-test-data/core": "4.4.0",
     "@commercetools-test-data/utils": "4.4.0",
-    "@commercetools/platform-sdk": "^4.0.0",
+    "@commercetools/platform-sdk": "^4.5.0",
     "@faker-js/faker": "^7.4.0"
   }
 }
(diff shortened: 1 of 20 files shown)


models/order/package.json
```diff
@@ -0,0 +1,33 @@
...
+    "@babel/runtime-corejs3": "^7.17.9",
+    "@commercetools-test-data/commons": "4.10.0",
+    "@commercetools-test-data/core": "4.10.0",
+    "@commercetools-test-data/utils": "4.10.0",
+    "@commercetools-test-data/line-item": "4.10.0",
+    "@commercetools-test-data/cent-precision-money": "4.10.0",
+    "@commercetools-test-data/customer-group": "4.10.0",
+    "@commercetools-test-data/cart": "4.10.0",
+    "@commercetools-test-data/cart-discount": "4.10.0",
+    "@commercetools/platform-sdk": "^4.5.0",
```
jaikamat, line 30: Does it matter that _just_ this module has the updated version? Do we need to update it across all modules?

jaikamat: the current SDK we were using on all our modules (4.0.0 I think?) was missing some features/fields on the Order type. I had to update the package here to access the new fields. Basically wondering if we need to update the SDK across all models or if it's fine here just being more up-to-date

jaikamat: Although shouldn't the `^` symbol download the most recent package within that semver anyway? Now I'm confused 🤔

jaikamat: I've reverted this in 35c3527 which didn't seem to cause any issues? 🤔

jaikamat: Ok, I think I just needed to nuke my `node_modules` folder and re-install to get the new version

stephsprinkle (approved):
😉

models/order/src/generator.ts
```diff
@@ -0,0 +1,85 @@
...
+    ),
+    discountCodes: [],
+    // TODO: unsupported by SDK types
+    // directDiscounts: [],
+    refusedGifts: fake(() => [CartDiscount.random()]),
+    cart: fake(() => Cart.random()),
+    quote: null,
+    custom: null,
+    paymentInfo: null,
+    locale: fake((f) => f.random.locale()),
```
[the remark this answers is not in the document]
jaikamat: Sounds good. Should we do what we do with the money models and go for something like `currencyCode: oneOf('EUR', 'USD')` here?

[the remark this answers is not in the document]
jaikamat: Addressed in a69a0f898de3095b6d1a2f32fd3514e797329ac9

Conversations with yourself are the best convos! 🦆

models/order/src/index.ts
```diff
@@ -0,0 +1,6 @@
+import * as OrderFromCartDraft from './order-from-cart-draft';
+export { OrderFromCartDraft };
+
+export { default as random } from './builder';
+export { default as presets } from './presets';
+export * from './types';
```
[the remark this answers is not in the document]
jaikamat: Nice catch! Addressed in 4eff666c334fe49562804be1c0285e9c8032f6bc

PR #219, opened 2023-03-03
FCT 185 sample data product model

tylermorrisford:
Adds the `product` model and draft, exposes the `product-variant` draft.

Please note a couple of questions about imports in the `generator.ts` files, and how to handle the `skus` field for graphql build on the `product` model 🙏


models/product/src/transformers.ts
```diff
@@ -0,0 +1,66 @@
...
+        .buildGraphql();
+
+      const taxCategoryRef: TReferenceGraphql = Reference.random()
+        .id(fields.productType.id)
+        .typeId('product-type')
+        .buildGraphql();
+
+      // TODO: doesn't exist on REST
+      // How should it be handled/built for graphql?
+      const skus: Array<String> = ['foo'];
```
jaikamat, line 53: One solution could be to add `skus` to a "Base" model type, then transform as needed to REST (by removing it), and to graphql (by keeping it)

tylermorrisford: fixed with [8d20c04](https://github.com/commercetools/test-data/pull/219/commits/8d20c0489493791ba794cf6cf49e02c03590ebdd) 💪

models/product/src/generator.ts
```diff
@@ -0,0 +1,31 @@
+import { ClientLogging, Reference } from '@commercetools-test-data/commons';
+import { fake, Generator, oneOf } from '@commercetools-test-data/core';
+import * as ProductCatalogData from '@commercetools-test-data/product-catalog-data'; // TODO: fix import
```
stephsprinkle, line 3: Don't need this comment anymore ✨

models/product/src/generator.ts
```diff
@@ -0,0 +1,31 @@
...
+    id: fake((f) => f.datatype.uuid()),
+    version: fake((f) => f.datatype.number()),
+    key: fake((f) => f.lorem.slug()),
+    productType: fake(() => Reference.random().typeId('product-type')),
+    masterData: fake(() => ProductCatalogData.random()),
+    taxCategory: fake(() => Reference.random().typeId('tax-category')),
+    state: fake(() => Reference.random().typeId('state')),
+    reviewRatingStatistics: null,
+    priceMode: oneOf(...Object.values(productPriceMode)),
+    skus: [],
```
jaikamat, line 23: Might be nice to see a stubbed out sku string value here

models/product/src/builder.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+import { TProduct, TProductGraphql } from './types';
+import * as Product from '.';
+
+describe('builder', () => {
+  it(
+    ...createBuilderSpec<TProduct, TProduct>(
+      'default',
+      Product.random(),
+      expect.objectContaining({
+        id: expect.any(String),
```
stephsprinkle, line 13: This one would have the `skus` array, correct?

models/product/src/types.ts
```diff
@@ -0,0 +1,29 @@
+import { TReferenceGraphql } from '@commercetools-test-data/commons';
+import type { TBuilder } from '@commercetools-test-data/core';
+import { Product, ProductDraft } from '@commercetools/platform-sdk';
+
+export type TProduct = Product & {
```
stephsprinkle, line 5: Do we need a REST type that doesn't have `skus`?

models/product/src/transformers.ts
```diff
@@ -0,0 +1,62 @@
...
+      'masterData',
+      'taxCategory',
+      'state',
+      'createdBy',
+      'lastModifiedBy',
+    ],
+  }),
+  graphql: Transformer<TProduct, TProductGraphql>('graphql', {
+    buildFields: [
+      'productType',
```
jaikamat, line 29: nit: `productType` here will still be a reference, even though it's documented as `ProductTypeDefinition`. The care level here is low as we're not using it, but it might confuse future consumers. Perhaps we can set `productType` to `null` via `replaceFields()` and add a note that the type doesn't exist (yet)?

I see this pattern reflected in the Orders PR I made [in this types file](https://github.com/commercetools/test-data/blob/35c352722c3f818645f075c49ddbe00c6503b650/models/order/src/types.ts), but admittedly I might have adhered to the documentation too strictly when the payoff is kind of minimal 🤷‍♂️

stephsprinkle (commented):
Looks really good, just a few comments

models/product/src/product-draft/generator.ts
```diff
@@ -0,0 +1,31 @@
...
+// https://docs.commercetools.com/api/projects/products#productdraft
+
+const generator = Generator<TProductDraft>({
+  fields: {
+    name: fake(() => LocalizedString.random()),
+    productType: fake(() => Reference.random().typeId('product-type')),
+    slug: fake(() => LocalizedString.random()),
+    key: fake((f) => f.lorem.slug()),
+    description: fake(() => LocalizedString.random()),
+    categories: fake(() => [Reference.random().typeId('categories')]),
```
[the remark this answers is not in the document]
stephsprinkle: 🎣

models/product/src/generator.ts
```diff
@@ -0,0 +1,31 @@
...
+const [getOlderDate, getNewerDate] = createRelatedDates();
+
+// https://docs.commercetools.com/api/projects/products#product
+
+const generator = Generator<TProduct>({
+  fields: {
+    id: fake((f) => f.datatype.uuid()),
+    version: fake((f) => f.datatype.number()),
+    key: fake((f) => f.lorem.slug()),
+    productType: fake(() => Reference.random().typeId('product-type')),
```
stephsprinkle, line 17: Should `productType`, `state`,  and `taxCategory` be the full representations and the fields build appropriate for the REST API (i.e. turned into references)?

models/product/src/transformers.ts
```diff
@@ -0,0 +1,62 @@
...
+        .buildGraphql();
+
+      const stateRef: TReferenceGraphql = Reference.random()
+        .id(fields.productType.id)
+        .typeId('state')
+        .buildGraphql();
+
+      const taxCategoryRef: TReferenceGraphql = Reference.random()
+        .id(fields.productType.id)
+        .typeId('product-type')
```
stephsprinkle, line 49:
```suggestion
        .typeId('tax-category')
```

models/product/src/builder.spec.ts
tylermorrisford: ⚡

models/product/src/generator.ts
tylermorrisford: 🔥

models/product/src/types.ts
tylermorrisford: YES! Should've been in the previous commit. Fixed with [7320171](https://github.com/commercetools/test-data/pull/219/commits/7320171b756be7ae071b24746474542739934d28)

models/product/src/generator.ts
tylermorrisford: fixed with [9ba2456](https://github.com/commercetools/test-data/pull/219/commits/9ba2456baf09e23376964b590f05214d1d34ecaf) (thanks for all your help Steph!!!)

models/product/src/generator.ts
stephsprinkle: I'm curious what your care amount is here. I'm assuming (and this is a big assumption) that in the GraphQL API that the skus are pulled from the product's variants and exposed via this field. Is that worthy of a transformer field or do we leave it here? My cares are so low.

models/product/src/transformers.ts
```diff
@@ -0,0 +1,65 @@
...
+    ],
+  }),
+  rest: Transformer<TProduct, TProductRest>('rest', {
+    buildFields: ['masterData', 'createdBy', 'lastModifiedBy'],
+    replaceFields: ({ fields }) => {
+      return {
+        ...fields,
+        productType: Reference.random().typeId('product-type').buildRest(),
+        taxCategory: Reference.random().typeId('tax-category').buildRest(),
+        // TODO: transform from field when state model is available
```
stephsprinkle, line 23: Do we need this comment anymore?

models/product/src/generator.ts
jaikamat: Yeah I agree, good point. Care amount also low, feels like there's not really a dire need to address it since we're not even using this model

models/product/src/transformers.ts
tylermorrisford: We said we were going to set it to null for now and build from the representation when the `state` model was completed, and then transform to a reference for graphql. But I may have not understood that properly.

stephsprinkle: You are doing the transform already - it's not going to change when the State model is available. The null check will remain because the field isn't required.

tylermorrisford: Ahh - ok. ditched it.

models/product/src/transformers.ts
tylermorrisford: Wondering if for time-boxing sake I should go back to this later. What do you think?

jaikamat: Yeah makes sense! Perhaps add a small note?

stephsprinkle (approved):
👏👏👏

models/product/src/transformers.ts
tylermorrisford: Strong - I'll surely forget if I don't. 🙏

stephsprinkle: I'm not sure I understand the issue here. In GraphQL, `productType` isn't a reference.

tylermorrisford: I removed the comment but I'll leave myself a note that this should possibly be revisited. This is where the graphql docs break down for me, I was making the assumption that for `productType` there was really only the `draft` and then once the entity was created it was called `ProductTypeDefinition` for whatever reason. Is that somewhat accurate?

jaikamat: This is no longer an issue now that the generator defines `productType` as not a reference, I missed that commit 👍

jaikamat (approved):

PR #220, opened 2023-03-06
PANGOLIN- 2462: locale options

valoriecarli:
AuLo e2es have been erring due to the randomness of locales vs the project language settings, which therefore fails in CI.

To correct this, the decision was made to employ of `locale: oneOf('en-US', 'de-DE', 'es-ES')` in the following existing models in which that property existed:
- Reviews
- Customers
- Cart 

is FCT ok with this?


jaikamat (approved):
Love it!

stephsprinkle (approved):

PR #221, opened 2023-03-06
WIP PANGOLIN-2385 - Change History preset pkg

valoriecarli:
**Goal**: Create a package preset (similar to [UI Kit presets](https://github.com/commercetools/ui-kit/tree/main/presets)) so that `audit-log` may import a single package that contains all necessary models.

⬇️ not set on the change-history preset name, but that will be worked out prior to publishing. @ByronDWall ... do you have a preference?

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -806,6 +806,8 @@ importers:
       '@commercetools-test-data/attribute-type': 4.10.0
       '@commercetools-test-data/category': 4.10.0
       '@commercetools-test-data/channel': 4.10.0
+      '@commercetools-test-data/commons': 4.10.0
+      '@commercetools-test-data/core': 4.10.0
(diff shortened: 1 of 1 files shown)


presets/change-history/src/index.ts
```diff
@@ -0,0 +1,12 @@
+export * as AttributeDefinition from '@commercetools-test-data/attribute-definition';
+export * as AttributeType from '@commercetools-test-data/attribute-type';
+export * as Category from '@commercetools-test-data/category';
+export * as Channel from '@commercetools-test-data/channel';
+export * as Commons from '@commercetools-test-data/commons';
+export * as Core from '@commercetools-test-data/core';
```
emmenko, line 6: Not sure if I would include these packages in this preset. In particular `core`, I don't think it should belong in here.

PR #223, opened 2023-03-15
PANGOLIN-2489- minor fix and index enhancements

valoriecarli:
This PR addresses 2 issues:
1.) Update a path in` models/price/src/price-draft/generator.ts` that was causing issues when imported into `aulo`.
2.) An attempt to expose remaining drafts, etc so they didn't have to discovered one at a time 😄


jaikamat (approved):
You got some eagle eyes!

This'll also help us when we have to resume work on our sample data import effort, so thank you in advance 🙏

tylermorrisford (approved):

PR #228, opened 2023-04-05
chore: remove unused lodash dep in some models

YahiaElTai:
This PR was initially intended to change the imports of lodash to `lodash/utility` to provide consistency with other repos  but then I found some instances where `lodash` was added into models and never used so I removed those as well. I hope that's okay 😅


jaikamat (approved):
🧹 🏅

PR #282, opened 2023-06-02
Change-History-Data - presets

valoriecarli:
To aid in efficiency & reliability for AuLo e2es, various presets & corresponding specs have been added:

- TaxRate
- TaxCategory
- ZoneRate
- ShippingRate
- ShippingMethod
- Customer
- Cart
- CartLineItem

Also fixed an oopsie in the CartLineItem generator & nulled Cutomer dob.


ByronDWall (approved):
tests pass, the presets make sense as far as I can tell, thanks!

PR #288, opened 2023-06-21
FCT-426 update shipping method names

tylermorrisford:
This PR renames the `ShippingMethodDraft` presets. 

Note: AFAIK these drafts are not used in other models and are only built in `mc-services` when the import takes place.

### DIFF
--- a/.changeset/hungry-dots-kick.md
+++ b/.changeset/hungry-dots-kick.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-test-data/shipping-method': minor
+---
+
+Update names of ShippingMethodDraft presets.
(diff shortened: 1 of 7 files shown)


stephsprinkle (approved):
Gosh I hope that the CoCo API auto handles the correct shipping method for the carts/orders. I searched the repo and couldn't find any references to these shipping methods 😅

tylermorrisford:
> Gosh I hope that the CoCo API auto handles the correct shipping method for the carts/orders. I searched the repo and couldn't find any references to these shipping methods 😅

Right 😬 ? 
Forgot a changeset, like you do

PR #289, opened 2023-06-22
PANGOLIN-2729: Location & AttributeLocalizedEnumValue models

valoriecarli:
This PR contains the following:

The addition of the `Location` model & a few presets. Based on the [documentation](https://docs.commercetools.com/api/projects/zones#representations), I went ahead and colocated Location to reside with Zone & organized the updated package accordingly.

The addition of the `AttributeLocalizedEnumValue` value, which is colocated w/in the `AttributeType` package. 

Please note that this only provides the _scaffolding_ for graphql. That portion will need more attention once we need it.

### DIFF
--- a/models/shipping-method/src/zone-rate/generator.ts
+++ b/models/shipping-method/src/zone-rate/generator.ts
@@ -1,5 +1,5 @@
 import { fake, Generator } from '@commercetools-test-data/core';
-import { Zone } from '@commercetools-test-data/zone/src';
+import { Zone } from '@commercetools-test-data/zone';
 import * as ShippingRate from '../shipping-rate';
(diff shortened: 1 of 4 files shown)


models/shipping-method/src/zone-rate/generator.ts
```diff
@@ -1,5 +1,5 @@
 import { fake, Generator } from '@commercetools-test-data/core';
-import { Zone } from '@commercetools-test-data/zone/src';
+import { Zone } from '@commercetools-test-data/zone';
```
tylermorrisford: 💪

tylermorrisford (approved):
LGTM!

PR #295, opened 2023-06-29
fix(models/cart): add shipping method to sample fashion data presets

stephsprinkle:
#### Summary

The Fashion sample data cart draft presets were missing shipping methods, which meant our orders were missing them as well.


jaikamat (approved):
🎊

PR #301, opened 2023-07-21
Added geolocation to channel draft

rajrdk:
This PR is intended to add geolocation to channel and channel draft.


models/channel/src/channel-draft/presets/with-inventory-supply-role-no-address.spec.ts
```diff
@@ -0,0 +1,14 @@
+import { roles } from '../../constants';
```
rajrdk, line 1: Added missing spec file

PR #302, opened 2023-07-24
fix(carts): add customer id to sample fashion data cart draft presets

stephsprinkle:
#### Summary
The FCT Fashion sample data is missing orders for customers in MC as we are only providing the customer's email and require the customer ID for that link to exist in MC.


jaikamat (approved):

models/cart/src/cart/cart-draft/presets/sample-data-fashion/sample-australia-01.ts
```diff
@@ -27,7 +27,7 @@ const shippingMethod = ShippingMethodDraft.presets.sampleDataFashion
   .usaAustralia()
   .build<TShippingMethodDraft>();
 
-const sampleAustralia01 = (customerId: string): TCartDraftBuilder =>
+const sampleAustralia01 = (customerId?: string): TCartDraftBuilder =>
```
tylermorrisford, line 30: <3 this

tylermorrisford (approved):

PR #305, opened 2023-07-26
PANGOLIN-2735 - Add `Stores` pkg with `Stores` & `ProductSelectionSettings` models

valoriecarli:
There is quite a bit happening here, and clearly room for improvement, but I think this is in a good place _for now_.

This PR consists of a new Stores package with the following models:
`ProductSelectionSettings`/ `ProductSelectionSettingsDraft`
`Store`/`StoreDraft`

Though the drafts are built up, the focus here was on the parent models - most notably the graphql transformations & their differing shapes/values.

@stephsprinkle / @jaikamat / @tylermorrisford  I have attempted to follow the graphql patterns FCT has established to make this easier for the next person. You do not need to review this, but please feel free if you'd like & let me know if you see a glaring mistake (or 3). 

Special thanks to @jaikamat for all the things.

### DIFF
--- a/.changeset/short-berries-rhyme.md
+++ b/.changeset/short-berries-rhyme.md
@@ -1,0 +1,6 @@
+---
+'@commercetools-test-data/product-selection': minor
+'@commercetools-test-data/store': minor
+---
+
+Add package `Store` with `Store` & `ProductSelectionSetting` models.


kterry1 (approved):
Awesome job!

ByronDWall (approved):
Looks good to me, thanks for putting in all this work!

PR #307, opened 2023-07-28
fix(carts): remove customer groups and update keys for fashion presets

stephsprinkle:
#### Summary
Found a couple bugs while testing the customer ID cart changes:
- Old cart keys
- Customer groups are not allowed on carts when providing customer ID as they are automagically added

Documentation for reference:
> [ResourceIdentifier](https://docs.commercetools.com/api/types#ctp:api:type:ResourceIdentifier) to the Customer Group of the Customer that the Cart belongs to. Used for [LineItem Price selection](https://docs.commercetools.com/api/carts-orders-overview#line-item-price-selection). It is automatically set if the Customer referenced in customerId belongs to a Customer Group. It can also be set explicitly when no customerId is present.


tylermorrisford (approved):

jaikamat (approved):
nice find!

PR #310, opened 2023-08-01
PANGOLIN-2973- `Payment` package creation w/  `PaymentInfoMethod` model

valoriecarli:
@tylermorrisford I hope this works for you. Apparently it doesn't care for scaffolding files 😄 , so I went ahead and created the `Payment` package & placed your `PaymentMethodInfo` model w/in it. Please feel free to build off of this, or push it. 

Screenshot 2023-08-01 at 11 28 30


tylermorrisford:
🚀 This is ahead of where I was, so THANK YOU! Going to add @stephsprinkle as a reviewer. Thank you so much @valoriecarli 🍾

.changeset/grumpy-dolphins-drop.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-test-data/payment': patch
```
stephsprinkle, line 2:
```suggestion
'@commercetools-test-data/payment': minor
```

stephsprinkle (commented):
This is one of those weird GraphQL models that has an `Input` type as a draft that doesn't exist in the REST API. ([AttributeDraft](https://github.com/commercetools/test-data/tree/main/models/product-variant/src/attribute/attribute-draft) is similar). 
image

models/payment/src/payment-method-info/payment-method-info-draft/transformers.ts
```diff
@@ -0,0 +1,28 @@
...
+const transformers = {
+  default: Transformer<TPaymentMethodInfoDraft, TPaymentMethodInfoDraft>(
+    'default',
+    {
+      buildFields: ['name'],
+    }
+  ),
+  rest: Transformer<TPaymentMethodInfoDraft, TPaymentMethodInfoDraft>('rest', {
+    buildFields: ['name'],
+  }),
```
jaikamat, line 16: Since there's no REST representation, we probably could omit this

models/payment/src/payment-method-info/payment-method-info-draft/transformers.ts
```diff
@@ -0,0 +1,28 @@
...
+  graphql: Transformer<TPaymentMethodInfoDraft, TPaymentMethodInfoDraftGraphQL>(
+    'graphql',
+    {
+      buildFields: ['name'],
```
jaikamat, line 20: `name` is an array of type `LocalizedStringItemInputType`, so we should probably transform from the LocalizedField we build (the data's there, it's just not the write shape I think - correct me if I'm wrong)

image

image

jaikamat (approved):
Few nits, but looks great! Nice work

models/payment/src/payment-method-info/payment-method-info-draft/transformers.ts
tylermorrisford: @jaikamat I haven't come across a transformation like this yet, is there an example you can think of? Or is it `LocalizedString.toLocalizedField(fields.name);`?

models/payment/src/index.ts
```diff
@@ -0,0 +1,5 @@
+// Export types
+export * from './payment-method-info/types';
+
+// Export models
+export * as Payment from './payment-method-info';
```
stephsprinkle, line 5: I just noticed this should almost definitely be 
```suggestion
export * as PaymentMethodInfo from './payment-method-info';
```

models/payment/src/payment-method-info/payment-method-info-draft/transformers.ts
stephsprinkle: @jaikamat It's genuinely unclear to me if we've ever run into this input type before. Can you think of an example?

stephsprinkle: The test for name looks like:
```
        name: expect.arrayContaining([
          expect.objectContaining({
            locale: expect.any(String),
            value: expect.any(String),
          }),
        ]),
```
Which, looks right to me? Am I missing something?

PR #318, opened 2023-08-14
FCT-456 PaymentStatus

tylermorrisford:
Creates the `paymentStatus` model and draft.


models/payment/src/payment-status/generator.ts
```diff
@@ -0,0 +1,14 @@
+import { Generator } from '@commercetools-test-data/core';
+import { TPaymentStatus } from './types';
+
+// https://docs.commercetools.com/api/projects/payments#paymentstatusdraft
+
+const generator = Generator<TPaymentStatus>({
+  fields: {
+    interfaceCode: null,
+    interfaceText: null,
+    state: null,
```
stephsprinkle, line 10: This would be a full-blown `State` model, which is then transformed into a `Reference` for the REST model and the `stateRef` field of GraphQL.

models/payment/src/payment-status/transformers.ts
```diff
@@ -0,0 +1,22 @@
+import { Reference } from '@commercetools-test-data/commons';
+import { Transformer } from '@commercetools-test-data/core';
+import type { TPaymentStatus, TPaymentStatusGraphql } from './types';
+
+const transformers = {
+  default: Transformer<TPaymentStatus, TPaymentStatus>('default', {
+    buildFields: ['state'],
+  }),
+  rest: Transformer<TPaymentStatus, TPaymentStatus>('rest', {
+    buildFields: ['state'],
```
stephsprinkle, line 10: `state` needs to be transformed into a `Reference` like line 17, I suppose you can do the same thing since the model doesn't exist yet.

models/payment/src/payment-status/payment-status-draft/generator.ts
```diff
@@ -0,0 +1,14 @@
+import { Generator } from '@commercetools-test-data/core';
+import { TPaymentStatusDraft } from '../types';
+
+// https://docs.commercetools.com/api/projects/payments#paymentstatusdraft
+
+const generator = Generator<TPaymentStatusDraft>({
+  fields: {
+    interfaceCode: null,
+    interfaceText: null,
+    state: null,
```
stephsprinkle, line 10: Here we can safely use the `Reference` since both REST and GraphQL expect it.

```suggestion
    state: fake(() => Reference.random().typeId('state')),
```

stephsprinkle (approved):

PR #334, opened 2023-09-13
Create sample-data-goodstore folder for store presets

lidiyacommercetools:
Creating a new folder to add zone information for the new sample data of the Good Store


tylermorrisford (approved):
This is a truly breathtaking PR right here

models/store/src/store/store-draft/presets/sample-data-goodstore
stephsprinkle, line 1: Did you intend to create this folder in zone drafts? Also wondering if we want to go with good-store vs goodstore. FWIW my care amount on this is low, we just need to be consistent for every model and this sets the precedent.

It would be nice if you could add an index.at that exports an empty presets constant. Here's an [example](https://github.com/commercetools/test-data/blob/main/models/zone/src/zone/zone-draft/presets/sample-data-fashion/index.ts), but it isn't empty.

stephsprinkle (commented):
Thank you, wonderful PM for your very first PR!!!!

models/store/src/store/store-draft/presets/sample-data-goodstore
lidiyacommercetools: I did mean to create it under zones!

stephsprinkle: One other thing, once you had the `index.ts` file in the sample data folder, you can then add the exported presets to the main `index.ts` file,  [like so](https://github.com/commercetools/test-data/blob/4246abd702b7b6ffbf30fc9a203d15b98ac1b4ec/models/zone/src/zone/zone-draft/presets/index.ts#L2). And that is basically the entirety of the "infrastructure" required to "drop in" presets.

stephsprinkle:
Closing in favor of #337

PR #345, opened 2023-09-16
feat(store): added store presets for goodstore data

lidiyacommercetools:
Added store presets to the good store data. Did not include the countries as the project only supports those two countries. Info seemed redundant. 

Mirrors orginal Good Store data. https://github.com/frontastic-developers/customer-thegoodstore/blob/master/getting-started/data/stores.json

### DIFF
--- a/models/store/src/store/store-draft/presets/sample-data-goodstore/the-good-store.ts
+++ b/models/store/src/store/store-draft/presets/sample-data-goodstore/the-good-store.ts
@@ -6,6 +6,5 @@ const the_good_store = (): TStoreDraftBuilder =>
   StoreDraft.presets
     .empty()
     .key('the-good-store')
-    .name(LocalizedString.presets.empty()['en-US']('The Good Store'))
-    .name(LocalizedString.presets.empty()['en-GB']('The Good Store'));
+    .name(LocalizedString.presets.empty()['en-US']('The Good Store')['en-GB']('The Good Store'));
 export default the_good_store;
--- a/models/store/src/store/store-draft/presets/sample-data-goodstore/the-good-store.ts
+++ b/models/store/src/store/store-draft/presets/sample-data-goodstore/the-good-store.ts
@@ -6,5 +6,10 @@ const the_good_store = (): TStoreDraftBuilder =>
   StoreDraft.presets
     .empty()
     .key('the-good-store')
-    .name(LocalizedString.presets.empty()['en-US']('The Good Store')['en-GB']('The Good Store'));
+    .name(
+      LocalizedString.presets
+        .empty()
+        ['en-US']('The Good Store')
+        ['en-GB']('The Good Store')
+    );
 export default the_good_store;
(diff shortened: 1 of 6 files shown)


models/store/src/store/store-draft/presets/sample-data-goodstore/the-good-store.ts
```diff
@@ -0,0 +1,11 @@
+import { LocalizedString } from '@commercetools-test-data/commons';
+import type { TStoreDraftBuilder } from '../../../types';
+import * as StoreDraft from '../../index';
+
+const the_good_store = (): TStoreDraftBuilder =>
+  StoreDraft.presets
+    .empty()
+    .key('the-good-store')
+    .name(LocalizedString.presets.empty()['en-US']('The Good Store'))
+    .name(LocalizedString.presets.empty()['en-GB']('The Good Store'));
```
tylermorrisford, line 10: Line 10 here will overwrite line 9; do we want both? We could do something like 
`.name(LocalizedString.presets.empty()['en-US']('The Good Store')['en-GB']('The Good Store'))`
to set the name for more than one language if that's what we need. Is [this](https://github.com/frontastic-developers/customer-thegoodstore/blob/master/getting-started/data/stores.json) what we want the end result to be?

lidiyacommercetools: Made the update. Thanks for highlighting

models/store/src/store/store-draft/presets/sample-data-goodstore/the-good-store.ts
```diff
@@ -0,0 +1,15 @@
+import { LocalizedString } from '@commercetools-test-data/commons';
+import type { TStoreDraftBuilder } from '../../../types';
+import * as StoreDraft from '../../index';
+
+const the_good_store = (): TStoreDraftBuilder =>
+  StoreDraft.presets
```
stephsprinkle, line 6: The original data includes `countries` and `languages`, do we want to include those here?

stephsprinkle (approved):

models/store/src/store/store-draft/presets/sample-data-goodstore/the-good-store.ts
```diff
@@ -0,0 +1,15 @@
+import { LocalizedString } from '@commercetools-test-data/commons';
+import type { TStoreDraftBuilder } from '../../../types';
+import * as StoreDraft from '../../index';
+
+const the_good_store = (): TStoreDraftBuilder =>
```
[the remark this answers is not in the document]
stephsprinkle: I agree that we should go with camel case

models/store/src/store/store-draft/presets/sample-data-goodstore/the-good-store.ts
lidiyacommercetools: Since we have only one store, and it supports all the languages and all the countries, is it necessary?

lidiyacommercetools: The API assumes all, unless stated otherwise

tylermorrisford (approved):

jaikamat (approved):

PR #347, opened 2023-09-18
feat: an-initial-exporter

phofmann:
DO NOT MERGE


stephsprinkle:
The presets in this PR need to be separated into their own PRs, verified against the source of truth, and unit tests need to be added.

phofmann:
No, the branch is important but the PR will never be applied

PR #362, opened 2023-09-25
feat(tax category): edit goodstore presets to include US

lidiyacommercetools:
Added US to the goodstore tax presets. The two US customers are from NY and NC, therefore only the two state taxes were added.


tdeekens (approved):
Noice 💯!

tylermorrisford (approved):
👯

tdeekens (approved):

PR #365, opened 2023-09-26
PANGOLIN-2385 - Include a package preset for  `change-history`

valoriecarli:
After a bit of rework, we have finally returned to this 🙌🏻 

This PR is to  create a package preset (similar to [UI Kit presets](https://github.com/commercetools/ui-kit/tree/main/presets)) so that `audit-log` may import a single package that contains only the necessary models (types, presets, etc) that we are consuming.

Once in place we will have 3 td dependencies instead of _18_:
`@commercetools-test-data/change-history`
`@commercetools-test-data/commons`
`@commercetools-test-data/core`

I have never attempted this before, so please take your time.

### DIFF
--- a/.changeset/dirty-rats-hug.md
+++ b/.changeset/dirty-rats-hug.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-test-data/change-history': patch
+---
+
+Add a `change-history` preset package specifically for the Audit Log application.


Sarah4VT (approved):

ByronDWall (commented):
The code looks good, but I have [thoughts](https://commercetools.slack.com/archives/C02HJ58DENM/p1696944646611289?thread_ts=1696880624.136279&cid=C02HJ58DENM) about whether or not we should implement this.

tldr, i personally think this is a premature/unnecessary abstraction, but if the rest of the team wants it im fine with implementing it

emmenko:
Thanks for the idea! 🤗

If the preset helps you and your team, then go ahead. If you can live without it, maybe we can leave it as-is for now.

> [!NOTE]
> For UI Kit we had the preset mostly for backwards compatibility reasons.
You also need to remember to include a new package to the preset.
>
> For UI Kit also we started thinking about other ways of publishing the packages (for example something similar to lodash).

PR #369, opened 2023-09-27
feat(carts): added cart presets for goodstore data

lidiyacommercetools:
DO NO REVEW
The PR is pending the merging of product and shipping method presets.


lidiyacommercetools:
PR included files that it was not meant to

PR #428, opened 2023-11-08
Remove status from custom view draft

Rhotimee:
### Summary

We don't need to have status in the customViewDraft. 

See the [CustomViewDraftDataInput](https://github.com/commercetools/merchant-center-services/blob/0edc1501facd3a24501dfacae65aed7197671168/packages/settings/lib/schemas/custom-view.graphql#L135) in the schema


emmenko (approved):
Good catch, thanks!

CarlosCortizasCT (approved):
Many thanks, Rotimi 🙇

PR #433, opened 2023-11-21
Pangolin 3184: staged quote model

ByronDWall:
This PR adds a new package for the `StagedQuote` and `StagedQuoteDraft` [models](https://docs.commercetools.com/api/projects/staged-quotes).

### DIFF
--- a/models/quote-request/src/transformers.ts
+++ b/models/quote-request/src/transformers.ts
@@ -133,7 +133,7 @@ const transformers = {
 
       const cartRef: TReferenceGraphql = Reference.presets.cartReference
         .cartReference()
-        .id(fields.cart?.id)
+        .id(fields.cart?.id || '')
         .typeId('cart')
         .buildGraphql();
 
@@ -150,7 +150,7 @@ const transformers = {
         storeRef: fields.store ? storeRef : undefined,
         stateRef: fields.state ? stateRef : undefined,
         cartRef: fields.cart ? cartRef : undefined,
-        businessUnitRef: fields.businessUnit ? businessUnitRef : undefined,
+        businessUnitRef,
         __typename: 'QuoteRequest',
       };
     },
--- a/models/quote-request/src/transformers.ts
+++ b/models/quote-request/src/transformers.ts
@@ -73,11 +73,6 @@ const transformers = {
         .id(fields.state?.id)
         .build<TReference<'state'>>();
 
-      const cart = Reference.presets.cartReference
-        .cartReference()
-        .id(fields.cart?.id)
-        .build<TReference<'cart'>>();
-
       const businessUnit = KeyReference.random()
         .typeId('business-unit')
         .key(fields.businessUnit?.key)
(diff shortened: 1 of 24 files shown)


models/staged-quote/src/types.ts
```diff
@@ -0,0 +1,42 @@
...
+
+export type TStagedQuoteGraphql = TStagedQuote & {
+  businessUnitRef: TReferenceGraphql | null;
+  customerRef: TReferenceGraphql | null;
+  quotationCartRef: TReferenceGraphql | null;
+  quoteRequestRef: TReferenceGraphql | null;
+  stateRef: TReferenceGraphql | null;
+  __typename: 'StagedQuote';
+};
+export type TStagedQuoteDraftGraphql = TStagedQuoteDraft;
```
jaikamat, line 37: Always a sigh of relief when they are symmetrical

models/staged-quote/src/types.ts
```diff
@@ -0,0 +1,42 @@
...
+
+// Rest
+export type TStagedQuoteRest = StagedQuote;
+export type TStagedQuoteDraft = StagedQuoteDraft;
+
+export type TStagedQuoteGraphql = TStagedQuote & {
+  businessUnitRef: TReferenceGraphql | null;
+  customerRef: TReferenceGraphql | null;
+  quotationCartRef: TReferenceGraphql | null;
+  quoteRequestRef: TReferenceGraphql | null;
```
jaikamat, line 33: image
Nit: `quotationCartRef` and `quoteRequestRef` aren't nullable

jaikamat (approved):
Spent time comparing types and they look good to me (with one nit)! My thinking is if those are good then the rest should flow from there. 

Nice work as always 🏆

models/staged-quote/src/types.ts
ByronDWall: this is a really good catch, I will go back and review the types in `quoteRequest` because I know that there are some non-nullable fields I did this to there as well, thanks!

ByronDWall: as it turns out, there were not any non-nullable errors in `QuoteRequest`, but I did find something else to fix, so, that's good I guess...

models/quote-request/src/transformers.ts
```diff
@@ -139,7 +133,7 @@ const transformers = {
 
       const cartRef: TReferenceGraphql = Reference.presets.cartReference
         .cartReference()
-        .id(fields.cart?.id)
+        .id(fields.cart?.id || '')
```
tylermorrisford, line 136: Ninja move!

tylermorrisford (approved):

models/quote-request/src/transformers.ts
jaikamat: Something like
```
      const cartRef: TReferenceGraphql | null = fields.cart
        ? Reference.presets.cartReference
            .cartReference()
            .id(fields.cart.id)
            .typeId('cart')
            .buildGraphql()
        : null;
```
could be more in line with the schema, that way we don't have to substitute empty strings and we know `id` is available

jaikamat: (full disclosure I did that [here](https://github.com/commercetools/test-data/blob/PANGOLIN-3184-staged-quote-model/models/order/src/transformers.ts#L73-L78))

ByronDWall: @jaikamat updated

jaikamat (approved):

PR #437, opened 2023-11-28
chore(business-unit): add typename fields to graphql transformers

ChristianMoll:
This adds the typename fields to the transformers of Division and Company. Thus enabling us to use it in our graphql based tests.


tdeekens (approved):

tdeekens:
Feel free to release if this blocks you. A release a day, keeps stale pipelines away.

PR #439, opened 2023-11-29
chore(business-unit): graphql transformers for Company and Division models

ChristianMoll:
This PR aims to replace the current scaffolding in the Company and Division models with the real implementation.

- Adds parentUnit field as null value in Company to prevent associated errors because of missing field
- Adds parentUnit and topLevelUnit to buildFields where applicable
- Adds ancestors and inheritedStores fields to graphQL transformer sprevent associated errors because of missing fields
- Replaces topLevelUnit field with itself for Company in graphQL transformer to match referenceExpansion
- Replaces parentUnit and topLevelUnit with Companies in Division graphQL transformer to match referenceExpansion

### DIFF
--- a/models/business-unit/src/company/transformers.ts
+++ b/models/business-unit/src/company/transformers.ts
@@ -3,16 +3,20 @@ import type { TBusinessUnit, TBusinessUnitGraphql } from '../types';
 
 const transformers = {
   default: Transformer<TBusinessUnit, TBusinessUnit>('default', {
-    buildFields: ['addresses', 'createdBy', 'lastModifiedBy'],
+    buildFields: ['addresses', 'createdBy', 'lastModifiedBy', 'topLevelUnit'],
   }),
   rest: Transformer<TBusinessUnit, TBusinessUnit>('rest', {
-    buildFields: ['addresses', 'createdBy', 'lastModifiedBy'],
+    buildFields: ['addresses', 'createdBy', 'lastModifiedBy', 'topLevelUnit'],
   }),
-
-  //only scaffolding provided at this time
   graphql: Transformer<TBusinessUnit, TBusinessUnitGraphql>('graphql', {
-    buildFields: [],
-    addFields: () => ({
+    buildFields: ['addresses', 'createdBy', 'lastModifiedBy'],
+    replaceFields: ({ fields }) => ({
+      ...fields,
+      topLevelUnit: {
+        ...fields,
+      },
+      ancestors: [],
+      inheritedStores: null,
       __typename: 'BusinessUnit',
     }),
   }),
--- a/models/business-unit/src/company/transformers.ts
+++ b/models/business-unit/src/company/transformers.ts
@@ -1,22 +1,19 @@
 import { KeyReference } from '@commercetools-test-data/commons';
 import { Transformer } from '@commercetools-test-data/core';
-import type { TBusinessUnit, TBusinessUnitGraphql } from '../types';
+import type { TCompany, TCompanyGraphql } from '../types';
 
 const transformers = {
-  default: Transformer<TBusinessUnit, TBusinessUnit>('default', {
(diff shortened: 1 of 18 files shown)


tylermorrisford (commented):
Would we want a test for the new graphql transformers?

ChristianMoll:
> Would we want a test for the new graphql transformers?

good catch! I will add some

models/business-unit/src/company/transformers.ts
```diff
@@ -3,16 +3,20 @@ import type { TBusinessUnit, TBusinessUnitGraphql } from '../types';
...
-
-  //only scaffolding provided at this time
   graphql: Transformer<TBusinessUnit, TBusinessUnitGraphql>('graphql', {
-    buildFields: [],
-    addFields: () => ({
+    buildFields: ['addresses', 'createdBy', 'lastModifiedBy'],
+    replaceFields: ({ fields }) => ({
+      ...fields,
+      topLevelUnit: {
+        ...fields,
```
tylermorrisford, line 16: I think this may just need `typeId: 'business-unit'` but I am not yet certain as to why 😿

tylermorrisford: I'm finding some other fields that may be missing from the graphql representation and should be added to the transformer: `parentUnitRef`, `topLevelUnitRef`, `storesRef`.

ChristianMoll: You're completely right, I missed those! I'll add them to the transformer and the tests

tylermorrisford:
This is looking good! Do you think it would make sense to create different types for company and division? I'm wondering if that would allow for better typing, or if it would not accomplish much. I will try and get you some more eyes on this 🙏

ChristianMoll:
> This is looking good! Do you think it would make sense to create different types for company and division? I'm wondering if that would allow for better typing, or if it would not accomplish much. I will try and get you some more eyes on this 🙏

Afaik the only difference is the parent unit / ref always being null/undefined for the parent unit and the unitType being Company or Division accordingly, so we would get a type that's a bit stricter for those properties and there are already Company and Division types from the SDK to use as the base. I think it's a good idea.

tylermorrisford:
> > > This is looking good! Do you think it would make sense to create different types for company and division? I'm wondering if that would allow for better typing, or if it would not accomplish much. I will try and get you some more eyes on this 🙏
> > 
> > 
> > Afaik the only difference is the parent unit / ref always being null/undefined for the parent unit and the unitType being Company or Division accordingly, so we would get a type that's a bit stricter for those properties and there are already Company and Division types from the SDK to use as the base. I think it's a good idea.
> 
> @tylermorrisford / @ChristianMoll I really like this and went back and forth on this issue it when I first built this. If we move forward with breaking it out, should we back and do the same for `draft` so there is symmetry (`TCompany`/ `TDivision` vs `TBusinesUnit`)?

I agree. Reduce the potential smell.

models/business-unit/src/types.ts
```diff
@@ -1,23 +1,56 @@
...
+export type TCreateCompanyDraftBuilder = () => TCompanyDraftBuilder;
+export type TCreateDivisionDraftBuilder = () => TDivisionDraftBuilder;
 //BusinessUnitDraftGraphql is only scaffolding at this time
-export type TBusinessUnitDraftGraphql = TBusinessUnitDraft & {
+export type TCompanyDraftGraphql = TCompanyDraft & {
+  __typename: 'BusinessUnitDraft';
+};
+export type TDivisionDraftGraphql = TDivisionDraft & {
   __typename: 'BusinessUnitDraft';
 };
```
tylermorrisford, line 25: Love this - thank you for all this work!

tylermorrisford (approved):

ChristianMoll:
@valoriecarli @tylermorrisford 
thank you very much for all the great feedback and suggestions!

PR #452, opened 2023-12-19
feat(presets): added presets to several entity drafts for use by audi…

Sarah4VT:
### Summary
Added presets to several entity drafts for use in audit log.

CategoryDraft.withNoParent
CustomerDraft.withNoDOB
InventoryEntryDraft.withNoSupplyChannel
ProductTypeDraft.withNoAttributes
ShoppingListDraft.withNoCustomer

### DIFF
--- a/.changeset/tricky-shrimps-guess.md
+++ b/.changeset/tricky-shrimps-guess.md
@@ -1,0 +1,8 @@
+---
+'@commercetools-test-data/inventory-entry': minor
+'@commercetools-test-data/product-type': minor
+'@commercetools-test-data/category': minor
+'@commercetools-test-data/customer': minor
+---
+
+Add presets to several entities for use by audit log
--- a/.changeset/tricky-shrimps-guess.md
+++ b/.changeset/tricky-shrimps-guess.md
@@ -1,5 +1,6 @@
 ---
 '@commercetools-test-data/inventory-entry': minor
+'@commercetools-test-data/shopping-list': minor
 '@commercetools-test-data/product-type': minor
 '@commercetools-test-data/category': minor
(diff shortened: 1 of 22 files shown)


.changeset/tricky-shrimps-guess.md
```diff
@@ -0,0 +1,8 @@
+---
```
Sarah4VT, line 1: Is there no publishing of the shopping list package yet? This didn't come up as a package when I created the changeset.

models/category/src/category-draft/presets/change-history-data/with-no-parent.spec.ts
```diff
@@ -0,0 +1,14 @@
+import type { TCategoryDraft } from '../../../types';
+import withNoParent from './with-no-parent';
+
+describe('Category with no parent', () => {
+  it('should return a category containing no parent value', () => {
+    const category = withNoParent().build<TCategoryDraft>();
+
+    expect(category).toEqual(
+      expect.objectContaining({
+        parent: undefined,
```
Sarah4VT, line 10: Do we want to test for the entire object or just the fields we are setting in the preset? I tried to look around but I feel like I found possibly cases of both.

Sarah4VT:
> I believe `models/shopping-list/src/shopping-list-draft/presets/index.ts` also needs to be updated to expose these properly:
> 
> ```
> import changeHistoryData from './change-history-data';
> const presets = {
>   changeHistoryData,
> };
> export default presets;
> ```
> 
> **As is** Screenshot 2023-12-27 at 15 18 56
> 
> **To be** Screenshot 2023-12-27 at 15 16 36

Thanks!!! I missed that. Good catch.

PR #467, opened 2024-01-31
feat: add model for organization and team

Rhotimee:


models/organization/src/team/types.ts
```diff
@@ -0,0 +1,25 @@
+import type { TBuilder } from '@commercetools-test-data/core';
+
+export type TTeam = {
+  id: string;
+  name: string;
+  members: []; // should be array of users. TODO: update when user model is ready
```
emmenko, line 6: We need to differentiate here between Rest models (`members` is a list of references) and GraphQL models (`members` is a list of CoCo Users).

models/organization/src/organization/types.ts
```diff
@@ -0,0 +1,31 @@
...
+  lastModifiedAt: string;
+};
+
+export type TOrganizationGraphql = TOrganization & {
+  __typename: 'Organization';
+};
+
+export type TOrganizationDraft = {
+  name: string;
+  owner: null; //Update type
```
emmenko, line 20: There is no `owner` in an Organization

models/organization/src/organization/types.ts
```diff
@@ -0,0 +1,31 @@
+import type { TBuilder } from '@commercetools-test-data/core';
+import type { TTeam } from '../team';
+
+export type TOrganization = {
+  id: string;
+  version: number;
+  name: string;
+  teams: TTeam[];
+  defaultClusters: null;
```
emmenko, line 9: I don't think we ever use this, I would omit it for now

emmenko (commented):
Thanks for working on this as well!

Just a few remarks for now as we wait for the `User` models to be finished.

models/organization/package.json
```diff
@@ -0,0 +1,26 @@
...
+    "access": "public"
+  },
+  "main": "dist/commercetools-test-data-organization.cjs.js",
+  "module": "dist/commercetools-test-data-organization.esm.js",
+  "files": ["dist", "package.json", "LICENSE", "README.md"],
+  "dependencies": {
+    "@babel/runtime": "^7.17.9",
+    "@babel/runtime-corejs3": "^7.17.9",
+    "@commercetools-test-data/core": "6.10.0",
+    "@commercetools-test-data/utils": "6.10.0",
```
kark, line 23:
```suggestion
    "@commercetools-test-data/core": "6.11.0",
    "@commercetools-test-data/utils": "6.11.0",
```

kark (commented):
~~Looks good to me 🙌 It seems like Nicola's comment (👆) regarding the owner in an organization hasn't been addressed yet.
I believe we're using a `organizationWithTeamsAndMembers` preset which e.g. defines an Administrators team. It would be beneficial to follow up on this when we integrate the User model~~

I'm sorry, now I've noticed you asked for a review of the other PR 🙈

kark (commented):
I have applied some changes to this PR.
- I've removed the `TeamDraft` as I couldn't find any trace of this type
- I've removed the `OrganizationDraft` but I up for bringing it back if you feel it's required (I believe [that's the type](https://github.com/commercetools/test-data/blob/2ce5141c25921ed6714b9916cd839d834de19619/graphql-types/src/generated/mc.ts#L314-L317) to model)
- I'm not 100% sure [I got this comment right](https://github.com/commercetools/test-data/pull/467#discussion_r1479361902)

kark:
> Is this PR ready for review? _I'm asking because it's still in Draft mode_

Yes, apologies. Thank you for the reminder

models/organization/src/team/transformers.ts
```diff
@@ -0,0 +1,16 @@
+import { Transformer } from '@commercetools-test-data/core';
+import type { TTeam, TTeamGraphql } from './types';
+
+const transformers = {
```
CarlosCortizasCT, line 4: Maybe what @emmenko meant was to return slightly different objects from here and keep just one Team model.

models/organization/README.md
```diff
@@ -0,0 +1,31 @@
+# @commercetools-test-data/organization
+
+This package provides the data model for `Organization` and `Team`.
```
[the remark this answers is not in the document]
emmenko: We try to keep entities grouped together when possible.

In this case, Teams are not very useful outside of Organizations so it makes sense to keep them together

models/organization/src/organization/generator.ts
```diff
@@ -0,0 +1,19 @@
...
+const [getOlderDate, getNewerDate] = createRelatedDates();
+
+const generator = Generator<TOrganization>({
+  fields: {
+    id: fake((f) => f.string.uuid()),
+    name: fake((f) => f.company.name()),
+    version: sequence(),
+    createdAt: fake(getOlderDate),
+    lastModifiedAt: fake(getNewerDate),
+    teams: fake(() => [Team.random()]),
```
emmenko, line 15: I would recommend to define the `Administrators` team as default

```suggestion
    teams: fake(() => [Team.random().name('Administrators')]),
```

models/organization/src/team/transformers.ts
emmenko: Yes correct. I think the Team model can use list of references by default but then in the graphql representation we need to change it to list of users.

I think we can use the `replaceFields` function for that. See maybe other models if they have a similar example.

emmenko (commented):
Thanks @kark  for picking this up! 🙌

models/organization/src/team/transformers.ts
kark: Thanks for the clarification and guidance! 🙌 I've implemented those suggestions.
Now, the default behavior is to create the REST model. For the GraphQL version, the `members` field is replaced, and `membersRef` and `__typename` fields are added in the GraphQL transformer.
Covered in 918296166d910994106e9a746876c0a81ca59f64

CarlosCortizasCT (approved):
Looks good to me 💯 

Thanks Kacper!

models/organization/src/team/types.ts
```diff
@@ -1,9 +1,16 @@
+import type { TReference } from '@commercetools-test-data/commons';
 import type { TBuilder } from '@commercetools-test-data/core';
 import type { TCoreTeam } from '@commercetools-test-data/graphql-types/src/generated/core';
 
 export type TTeam = Omit<TCoreTeam, '__typename'>;
 
 export type TTeamGraphql = TCoreTeam;
 
-export type TTeamBuilder = TBuilder<TTeam>;
+export type TTeamRest = {
```
emmenko, line 9: Nit: I would name this `TTeam` and the GraphQL representation `TTeamGraphql`. No other types should be needed.

models/organization/src/team/generator.ts
```diff
@@ -1,14 +1,12 @@
 import { Reference } from '@commercetools-test-data/commons';
 import { fake, Generator } from '@commercetools-test-data/core';
-import { User } from '@commercetools-test-data/user';
-import type { TTeam } from './types';
+import type { TTeam, TTeamRest } from './types';
 
-const generator = Generator<TTeam>({
+const generator = Generator<TTeam | TTeamRest>({
```
emmenko, line 5: We should use only the default type 

```suggestion
const generator = Generator<TTeam>({
```

models/organization/src/team/builder.spec.ts
```diff
@@ -1,62 +1,77 @@
...
-      id: expect.any(String),
-      typeId: 'user',
-      ...(args?.isGraphql ? { __typename: 'Reference' } : {}),
-    }),
-  ]);
-
 describe('builder', () => {
   it(
-    ...createBuilderSpec<TTeam, TTeam>(
+    ...createBuilderSpec<TTeam | TTeamRest, TTeamRest>(
```
emmenko, line 9: Based on the previous suggestions:

```suggestion
    ...createBuilderSpec<TTeam, TTeam>(
```

models/organization/src/team/transformers.ts
```diff
@@ -1,13 +1,21 @@
...
+    buildFields: ['members'],
   }),
   graphql: Transformer<TTeam, TTeamGraphql>('graphql', {
-    buildFields: ['members', 'membersRef'],
-    addFields: () => ({
+    buildFields: ['members'],
+    replaceFields: ({ fields }) => ({
+      ...(fields as TTeam),
+      members: [User.random().buildGraphql()],
+      membersRef: [Reference.random().typeId('user').buildGraphql()],
```
emmenko, line 18: 💯

emmenko (commented):
Thanks for the fixes! This looks good to me now, I left some final suggestions.

🙌

models/organization/src/team/types.ts
kark: Oh, yes. These are great suggestions, applied in 2aeb66a05f07c24c1bc179e2bfae8c407fdcbd53

emmenko (approved):
💯

PR #473, opened 2024-02-02
chore: remove pangolin from models catch-all ownership

wrsenn:
The Pangolin team isn't responsible for shared tooling and can't reliably devote time to this. However, if the goal is to properly assign ownership to the right team when changes are made, it seems like that should be directed by tech leadership with more granular code ownership assignments. Ideally, any time a new model is added or an existing, unowned model is updated, `CODEOWNERS` should be updated to reflect the correct team at that time, and we can't really make that call.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -19,7 +19,7 @@
 /codegen* @commercetools/shield-team-fe
 
 # Models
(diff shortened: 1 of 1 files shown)


tdeekens (approved):
Fair. We can also assign it to Test Automation Engineering.

PR #444, opened 2023-12-01
feat(graphql drafts): attempt at removing __typename from transformed fields for graphql drafts

ByronDWall:
When transforming fields for graphql draft models, built fields will add a `__typename`, which is not a valid field to send in a draft and makes API's very sad and send us a `400`.  

This is the beginnings of an attempt to add a transform option `isGraphqlDraft` that would remove `typename` from any built fields if passed.

PR #474, opened 2024-02-05
FCT 598: Omit `__typename` field when building graphql draft models

ByronDWall:
When transforming fields for graphql draft models, built fields will add a `__typename` field, which is not a valid field to send in a draft and makes API's very sad and send us a 400.

This PR implements an `isGraphqlDraft` option that can be passed to the `Transformer` function.  When this option is passed, all `__typename` fields are removed from the transformed model so that the model can be used directly as a draft.

For a more in depth walkthrough and explanation of this code, see [this previous PR](https://github.com/commercetools/test-data/pull/444).

### DIFF
--- a/.changeset/unlucky-phones-repair.md
+++ b/.changeset/unlucky-phones-repair.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-test-data/core': minor
+---
+
+"fix(graphql drafts): omit \_\_typename field from built graphql draft models when isGraphqlDraft option is passed to transformer"
--- a/core/src/builder.spec.ts
+++ b/core/src/builder.spec.ts
@@ -506,6 +506,49 @@ describe('building', () => {
               ],
             });
           });
+          it('should omit __typename field when building nested builders and isGraphqlDraft is true', () => {
+            const teamTransformers = {
+              graphql: Transformer<TestTeam, TestTeam>('graphql', {
+                buildFields: ['users'],
+                isGraphqlDraft: true,
+              }),
+            };
+            const userTransformers = {
+              graphql: Transformer<
+                TestExpandedUserReference,
+                TestExpandedUserReferenceGraphql
+              >('graphql', {
+                addFields: () => ({
+                  __typename: 'User',
+                }),
+                isGraphqlDraft: true,
+              }),
+            };
+            const userBuilder1 = Builder<TestExpandedUserReference>({
+              transformers: userTransformers,
+            }).name('My name');
+            const userBuilder2 = Builder<TestExpandedUserReference>({
+              transformers: userTransformers,
+            }).name('My other name');
+            const built = Builder<TestTeam>({
+              transformers: teamTransformers,
+            })
+              .id('my-id')
+              .users<TestExpandedUserReference>([userBuilder1, userBuilder2])
+              .buildGraphql<TestTeam>();
+
+            expect(built).toEqual({
+              id: 'my-id',
+              users: [
+                {
+                  name: 'My name',
+                },
+                {
+                  name: 'My other name',
+                },
+              ],
+            });
+          });
         });
       });
     });
--- a/core/src/helpers.ts
+++ b/core/src/helpers.ts
@@ -28,6 +28,33 @@ const upperFirst = (value: string): string =>
 const lowerFirst = (value: string): string =>
   value.charAt(0).toLowerCase() + value.slice(1);
 
+// eslint-disable-next-line  @typescript-eslint/no-explicit-any
+function deleteKeyFromObject(inputObject: any, keyToDelete: string) {
+  for (let [currentObjectKey, currentObjectValue] of Object.entries(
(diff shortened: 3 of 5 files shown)


jaikamat:
Is there any use-case for providing the draft `__typename` from a consumer perspective? It's obviously a field that's needed to conform to a GraphQL schema since `__typename` is necessary, but could this functionality be made automatic?

I'm thinking about the `types.ts` files in which we make efforts to describe accurate `__typename`s based on the schema, wondering if we should always assume that for draft entities, `__typename` will be omitted?

tdeekens (commented):
Thanks. This is really not ideal! Wondering if instead of leaking a "magic" boolean into the root type of model, we could add a `removeFields` to any draft model [here](https://github.com/commercetools/test-data/blob/main/models/associate-role/src/associate-role-draft/transformers.ts#L12). It's a bit more work but then better located maybe?

ByronDWall:
> Thanks. This is really not ideal! Wondering if instead of leaking a "magic" boolean into the root type of model, we could add a `removeFields` to any draft model [here](https://github.com/commercetools/test-data/blob/main/models/associate-role/src/associate-role-draft/transformers.ts#L12). It's a bit more work but then better located maybe?

@tdeekens Im certainly not against it, but I will have to look into how that would work a little more to understand it completely from my end

tdeekens:
Thanks. I'd value that. We'd mainly add a `removeFields` to every transformer of each draft type in the graphql transformer. It's more repetitive but should work to my memory.

ByronDWall:
> Thanks. I'd value that. We'd mainly add a `removeFields` to every transformer of each draft type in the graphql transformer. It's more repetitive but should work to my memory.

@tdeekens We already have a follow-up pr planned that will do essentially what you are asking, but the approach you are suggesting will not remove `__typename` values from _nested_ transformed fields (eg a `productType` reference in products) since `fieldsToRemove` runs a `delete` [only on fields that match in the top-level object](https://github.com/commercetools/test-data/blob/65dbb7d950a1a05133f69141208c03b8893bf83e/core/src/transformer.ts#L77).

Deleting nested fields with `removeFields` seems like it could introduce unwanted and unexpected behavior.  Which, thinking back on the original draft PR that I originally put up a couple months ago, this is a large reason why I went with making a specific "magic" boolean option on the root transformer type, since graphql drafts are a bit of a different usecase.

jaikamat:
@tdeekens This makes me wonder if there's value in typing `__typename` for drafts and creating it at all? Should we remove all `__typename` declarations from `types.ts` and not add them in the GraphQL transforms, for draft entities?

tdeekens:
> @tdeekens This makes me wonder if there's value in typing `__typename` for drafts and creating it at all? Should we remove all `__typename` declarations from drafts in `types.ts` and not add them in the GraphQL transforms, for draft entities?

Yes. That would be another more native option to test-data. If we prevent it from being added at all that solve the root cause nicely.

jaikamat:
> > @tdeekens This makes me wonder if there's value in typing `__typename` for drafts and creating it at all? Should we remove all `__typename` declarations from drafts in `types.ts` and not add them in the GraphQL transforms, for draft entities?
> 
> Yes. That would be another more native option to test-data. If we prevent it from being added at all that solve the root cause nicely.

Honest question - if we set the expectation that draft GraphQL entities never have `__typename` built, can we carry this over to non-draft models? This means that consumers won't have the ability to assert against `__typename` in testing, but I wonder how problematic this is.

Seems to me that if we don't build `__typename` for draft models, then they're already not GraphQL schema-compliant.

We could definitely audit instances of `__typename`, remove them, and greatly simplify some of these issues (especially for dual use-case models like `Reference` where there is no draft model)

jaikamat:
> Overall this is the crux. As per GraphQL spec __typename is only used for Object, Interface, or Union. Not an Input type - what is passed with a mutation.
> 
> If an Model was only used for an input type in a GraphQL schema. Then we could safely remove it. If however, drafts are also used for "unsaved" entities there is an issue. As in that case it would be a mocked response and flow into e.g Apollo's cache.
> 
> If drafts are used as responses to GraphQL queries?
> 
> Does this make sense?

Toooootally makes sense now given that context, as draft models are effectively input types. So seems like the path forward is to create draft models for all "dual-use-case" models (like `Reference`), use them where necessary, and never specify a `__typename` to be built?

tdeekens:
> Toooootally makes sense now given that context, as draft models are effectively input types. So seems like the path forward is to create draft models for all "dual-use-case" models (like Reference), use them where necessary, and never specify a __typename to be built?

Yes to my current understanding. For context: we had a time in which a draft was a preset of an entity in which we unset the __typename using eg .__typename(undefined). Still as drafts and regular entities at times are structurally too different we decided to split them in two models. Given that, and the clear input oriented use case of drafts, we might just never specify a typename to be build. 

Unless of course there are other valid and good reasons to keep it.

PR #476, opened 2024-02-06
feat(order): add orderFromQuoteDraft model

tylermorrisford:
Adds the `OrderFromQuoteDraft` [model](https://docs.commercetools.com/api/projects/orders#orderfromquotedraft) to the Orders package.

### DIFF
--- a/models/order/src/order-from-quote-draft/generator.ts
+++ b/models/order/src/order-from-quote-draft/generator.ts
@@ -1,0 +1,26 @@
+import { Reference } from '@commercetools-test-data/commons';
+import {
+  fake,
+  Generator,
+  oneOf,
+  sequence,
+} from '@commercetools-test-data/core';
+import { orderState, paymentState, shipmentState } from '../constants';
+import { TOrderFromQuoteDraft } from '../types';
+
+// https://docs.commercetools.com/api/projects/orders#orderfromquotedraft
+
+const generator = Generator<TOrderFromQuoteDraft>({
+  fields: {
+    version: sequence(),
+    quote: fake(() => Reference.random().typeId('quote')),
+    quoteStateToAccepted: fake(() => false),
+    orderNumber: fake((f) => String(f.number.int({ min: 100000 }))),
+    paymentState: oneOf(...Object.values(paymentState)),
+    shipmentState: oneOf(...Object.values(shipmentState)),
+    orderState: oneOf(...Object.values(orderState)),
+    state: fake(() => Reference.random().typeId('state')),
+  },
+});
+
+export default generator;
--- a/models/order/src/order-from-quote-draft/generator.ts
+++ b/models/order/src/order-from-quote-draft/generator.ts
@@ -14,7 +14,7 @@ const generator = Generator<TOrderFromQuoteDraft>({
   fields: {
     version: sequence(),
     quote: fake(() => Reference.random().typeId('quote')),
-    quoteStateToAccepted: fake(() => false),
+    quoteStateToAccepted: fake((f) => f.datatype.boolean()),
     orderNumber: fake((f) => String(f.number.int({ min: 100000 }))),
     paymentState: oneOf(...Object.values(paymentState)),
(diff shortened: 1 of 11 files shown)


models/order/src/order-from-quote-draft/generator.ts
```diff
@@ -0,0 +1,26 @@
...
+import { orderState, paymentState, shipmentState } from '../constants';
+import { TOrderFromQuoteDraft } from '../types';
+
+// https://docs.commercetools.com/api/projects/orders#orderfromquotedraft
+
+const generator = Generator<TOrderFromQuoteDraft>({
+  fields: {
+    version: sequence(),
+    quote: fake(() => Reference.random().typeId('quote')),
+    quoteStateToAccepted: fake(() => false),
```
stephsprinkle, line 17: Why not a fake boolean that could be true?

models/order/src/order-from-quote-draft/presets/empty.ts
```diff
@@ -0,0 +1,13 @@
+import type { TOrderFromQuoteDraftBuilder } from '../../types';
+import OrderFromQuoteDraft from '../builder';
+
+const empty = (): TOrderFromQuoteDraftBuilder =>
```
stephsprinkle, line 4: Bless you for adding this preset

stephsprinkle (approved):

models/order/src/order-from-quote-draft/generator.ts
ByronDWall, line 17:
```suggestion
    quote: Reference.random().typeId('quote'),
    quoteStateToAccepted: false,
```

I don't think you need to wrap directly declared values in `fake` if you arent using `faker` (eg `f.number...` or whatever)

ByronDWall: you can specify it as true when you call the function, it's probably not a good idea to default all drafts to generate accepted quotes?

ByronDWall (approved):
Overall lgtm, just one small nit that I have very little care about either way

models/order/src/order-from-quote-draft/generator.ts
tylermorrisford: Docs say the default is `false` so I rolled with it, but I did think about using the fake boolean, pinky swear.

models/order/src/order-from-quote-draft/presets/empty.ts
tylermorrisford: 🙏

models/order/src/order-from-quote-draft/generator.ts
tylermorrisford: strong point, tbh.

tylermorrisford: Fixed in [a663e4d](https://github.com/commercetools/test-data/pull/476/commits/a663e4d6e812397deb5768129d16ba746b65fc6c)

models/order/src/order-from-quote-draft/generator.ts
```diff
@@ -0,0 +1,26 @@
...
+} from '@commercetools-test-data/core';
+import { orderState, paymentState, shipmentState } from '../constants';
+import { TOrderFromQuoteDraft } from '../types';
+
+// https://docs.commercetools.com/api/projects/orders#orderfromquotedraft
+
+const generator = Generator<TOrderFromQuoteDraft>({
+  fields: {
+    version: sequence(),
+    quote: fake(() => Reference.random().typeId('quote')),
```
tdeekens, line 16: Should we merge @jaikamat PR first and make this a ReferenceDraft which he introduced over there?

tylermorrisford: Sounds good!

tylermorrisford: @tdeekens It is done ✅ 🎸

jaikamat (approved):

PR #478, opened 2024-02-07
[Breaking]: architectural change to omit draft model `__typename` creation

jaikamat:
## Summary

This PR makes sweeping changes to draft models, and introduces a proposed ADR we can agree on going forward.

# The problem

Note that the changes in this PR only apply to draft models, not base models.

While `__typename` is technically part of the GraphQL schema for draft/input types, using it in practice has proven challenging. When we build draft GraphQL models and attempt to directly submit them as variables via the CoCo API, we are met with `400` errors, because we build `__typename` on many draft models internally. 

This has led to our team having to shim `omitDeep` functions in our sample data creation resolvers to strip `__typename`, and casting to appropriate types, which is neither type safe nor ideal. I suspect that some consumers using draft models will have to do the same.

# Proposed solution
@ByronDWall had done great work trying to solve this [here](https://github.com/commercetools/test-data/pull/474), however that raised questions about the very purpose of draft models. 

We learned that models were originally designed without drafts, but as testing necessitated mocking requests, drafts were split out to represent input types (and as many of you know, GraphQL drafts may deviate wildly from their base REST representations, so splitting them out proved very maintainable).

Therefore, **we're setting the precedent** here that draft GraphQL models should adhere to their preferred utilitarian shape (which excludes `__typename`), rather than their strict schema representation. After all, this is a test data repository, and our form should follow function.

There was one lingering problem with this: `Reference`, `KeyReference`,  and `LocalizedString` did not have draft representations, and always included `__typename` when built for GraphQL. They essentially had a "dual use case". If we allowed them to be built as-is, our live API requests would still fail.

Therefore, they have been extended to include their respective `*Draft` representations (which exclude `__typename`), and all points of implementation in draft models have been edited to use them.

Further, we have removed `__typename` from those draft models that built it.

Tests and snapshots have been updated accordingly.

## TL;DR
Draft GraphQL models no longer have `__typename`. Please indicate if this will have downstream complications for your team. We are considering this a breaking change


jaikamat:
image
@commercetools/first-contact-team-fe No draft preset snapshot tests contain `__typename`

models/business-unit/src/types.ts
```diff
@@ -17,12 +17,8 @@ export type TDivisionDraftBuilder = TBuilder<TDivisionDraft>;
 export type TCreateCompanyDraftBuilder = () => TCompanyDraftBuilder;
 export type TCreateDivisionDraftBuilder = () => TDivisionDraftBuilder;
 //BusinessUnitDraftGraphql is only scaffolding at this time
-export type TCompanyDraftGraphql = TCompanyDraft & {
-  __typename: 'BusinessUnitDraft';
```
jaikamat, line 21: For most of the models in this PR (save for `Reference`, `KeyReference`, and `LocalizedString`), this was the relevant change. Remove `__typename` from `types.ts`, the generator, and ensure the type was safe all the way down to tests

docs/architecture-decisions/0002-draft-model-structure.md
```diff
@@ -0,0 +1,28 @@
+# 2. Exclude `__typename` from Draft GraphQL Models
```
jaikamat, line 1: Highlighting the ADR here for others to review 🔍

docs/architecture-decisions/0001-record-architecture-decisions.md
```diff
@@ -0,0 +1,43 @@
+# 1. Documentation: Record architecture decisions
```
tdeekens, line 1: I can't say how much I value this. I've thought last night "I'll add the ADRs once the PR is in". Then I view it and you added it already.

tdeekens (commented):
🏩 The changes look very sane. I value taking the time taken for this approach! Especially the ADR addition. 

The main question I guess now: breaking or not. I would rather be safe and make it a breaking release. WDYT? Consumers can still add a typename to drafts by adding `__typename('foo')` I suppose?

docs/architecture-decisions/0001-record-architecture-decisions.md
jaikamat: Gotchu fam! ✨

jaikamat:
> 🏩 The changes look very sane. I value taking the time taken for this approach! Especially the ADR addition.
> 
> The main question I guess now: breaking or not. I would rather be safe and make it a breaking release. WDYT? Consumers can still add a typename to drafts by adding `__typename('foo')` I suppose?

Definitely comfortable labeling this as a breaking release considering the number of presets and files it touches. How do I go about doing that, is it part of the changeset?

To be honest consumers will likely have some difficulty tacking `__typename` on, especially for deeply-nested models, but that's the tradeoff. If teams are relying on `__typename` in draft GraphQL model tests, I'd be curious to see how widespread of an issue it is - but my suspicions tell me it's minimal, if at all.

emmenko (approved):
Approving the approach, the documentation and the effort put into this! Many thanks ❤️

I won't be reviewing every single change though, I trust your abilities 😉

jaikamat:
@tdeekens see cc0edf3caa27603c446f623f7513e1165be82a36 for the changeset 👍

tdeekens (approved):
Super duper. So proud how far test-data came from a small idea to things falling into place.

stephsprinkle (approved):
This is a huge improvement 👏

jaikumar-tj (approved):
👏 Thanks for such a massive work

kterry1 (approved):
Awesome job Jai!

tdeekens:
:shipit:

PR #488, opened 2024-02-23
PANGOLIN-3515 - CH MoneyDraft & Payment presets

valoriecarli:
The following is covered in this PR:

- Adjustment to the CH PaymentDraft preset `withUsdCurrencyCode` to accommodate recent fixes to to `PaymentStatus` of the last major release.
- Further narrowing down of a the `withUsdCurrencyCodeCentMax` preset for CH `MoneyDraft` 
- Pull the `withUsdCurrencyCodeCentMax` preset into `PaymentDraft.amountPlanned`, as this was also causing an `Int cannot represent non 32-bit signed integer value` error (discovered once `Transaction.amount `was settled).

### DIFF
--- a/.changeset/thin-rivers-film.md
+++ b/.changeset/thin-rivers-film.md
@@ -1,0 +1,6 @@
+---
+'@commercetools-test-data/commons': patch
+'@commercetools-test-data/payment': patch
+---
+
+Adjustments made to CH MoneyDraft & PaymentDraft presets.


Sarah4VT:
Approving but do want to see the follow up happen in https://commercetools.atlassian.net/browse/PANGOLIN-3488 as I suspect we can get rid of the withUsdCurrencyCodeCentMax if we update the MoneyDraft generator to only generate values that can possibly be stored in the database and be accepted by the coco api. If we know this is a needed global fix, we shouldn't contain it to just our presets.

Sarah4VT (approved):

jmcreasman (commented):
🚀

PR #499, opened 2024-02-28
PANGOLIN-3529 key reference fix

valoriecarli:
minor fix to the recently updated BU generator fields (so it wouldn't pass `unknown`)
updated corresponding specs that allowed this to pass.


ByronDWall (approved):
lgtm

tylermorrisford (approved):
🕺

PR #509, opened 2024-02-28
fix(presets/b2b): add channels to line items

stephsprinkle:
For cart creation purposes (and standalone prices)


ByronDWall (approved):
its like I just watched you do this

tylermorrisford (approved):
changeset?

PR #511, opened 2024-02-29
chore(td): fix quote presets

jaikamat:


ByronDWall (approved):
🔪

tylermorrisford (approved):
watched the magic happen with mine own eyes

PR #521, opened 2024-03-04
[B2B Sample Data]: adding channel specific inventories

NickDevG:
##  [B2B Sample Data]: adding channel specific inventories

Adding the inventory entries with channels to allow inventory tracking by channel. The generic ones as they are not used on the project were removed.


models/inventory-entry/src/inventory-entry-draft/presets/sample-data-b2b/sku-a789-bc-2015-default.ts
```diff
@@ -0,0 +1,28 @@
...
+  .a789Bc01()
+  .build<TProductVariantDraft>();
+
+const skuA789Bc2015Default = (): TInventoryEntryDraftBuilder =>
+  InventoryEntryDraft.presets
+    .empty()
+    .key(`${a789Bc01.sku}-${defaultWarehouse.key}`)
+    .sku(a789Bc01.sku!)
+    .quantityOnStock(1000)
+    .channel({ typeId: 'channel', key: defaultWarehouse.key });
```
stephsprinkle, line 26: These should be `KeyReferenceDraft.presets.channel().key(defaultWarehouse.key)`. And like the linter pointed out, `channel` doesn't exist, perhaps this should be `supplyChannel`?

stephsprinkle (changes requested):

models/inventory-entry/src/inventory-entry-draft/presets/sample-data-b2b/sku-a789-bc-2015-default.ts
NickDevG: Absolutely right. I updated the files

ByronDWall (approved):
Thanks for taking the time to do all this @NickDevG! 

the schema looks good to me, and the keys are unique as far as I can tell, approved

stephsprinkle:
Please add a changeset!

PR #526, opened 2024-03-07
PANGOLIN-35778- CH presets

valoriecarli:
As the title indicates.

Adding presets to prevent the `ProductDiscount` spec from nuking the `Product` suite.


kterry1 (approved):
Great job!

PR #528, opened 2024-03-11
Reduce b2b sample data

NickDevG:
## Reduce b2b sample data

- Update carts to use remaining sku's
- Remove product variants
- Remove products
- Remove inventory entries
- Remove standalone prices
- Remove categories


jaikamat (approved):
Checks pass, we will try this on our end - you're a champion of the highest order!

ByronDWall (approved):
checks passed, we will try this on our end and update if/as-needed

huge thanks @NickDevG we really appreciate it!

PR #536, opened 2024-03-26
FCT-783: Add Product Selections to Stores

ByronDWall:
- generates `productSelectionSettingsDraft` presets for both b2b product-selections
- references appropriate `PSSD`'s in b2b `storeDraft` presets
- creates `productSelection` keyReference preset
- adds and updates appropriate tests

### DIFF
--- a/models/store/src/store/store-draft/presets/sample-data-b2b/de-fr-uk.ts
+++ b/models/store/src/store/store-draft/presets/sample-data-b2b/de-fr-uk.ts
@@ -6,6 +6,10 @@ import {
   KeyReferenceDraft,
   LocalizedStringDraft,
 } from '@commercetools-test-data/commons';
+import {
+  ProductSelectionSettingDraft,
+  type TProductSelectionSettingDraft,
+} from '../../../../product-selection-setting/index';
 import type { TStoreDraftBuilder } from '../../../types';
 import * as StoreDraft from '../../index';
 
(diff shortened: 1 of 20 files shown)


tylermorrisford (approved):
This looks really good to me - agree on the naming.

jaikamat (approved):

models/store/src/store/store-draft/presets/sample-data-b2b/de-fr-uk.ts
```diff
@@ -39,6 +48,7 @@ const deFrUk = (): TStoreDraftBuilder =>
     ])
     .supplyChannels([
       KeyReferenceDraft.presets.channel().key(euWarehouseChannel.key),
-    ]);
+    ])
+    .productSelections([defaultProductSelectionSetting]);
```
stephsprinkle, line 52: Does this need to be inside a `faker`? I feel like we've seen that before?

stephsprinkle (commented):
Does `productSelection` need to be added to the store draft transformer?

models/store/src/store/store-draft/presets/sample-data-b2b/de-fr-uk.ts
ByronDWall: how is it different than `supplyChannels`, since it's pre-built?  I can certainly put it in a `fake` block tho

ByronDWall: thinking about this more, if we wanted to put a random product selection into the default (`.random()`) `StoreDraft` defined in `store-draft/generator.ts`, we would specify that in a faker block.

but I don't believe we need a faker block for a one-off pre-built value in a preset

PR #543, opened 2024-03-28
[B2B sample data]: adding missing differences

NickDevG:
## Adding missing differences

Looking for the remaining differences between the sample data and the auto-generated data from the MC


ByronDWall (approved):
As always, we truly appreciate it @NickDevG !

models/commons/src/address/address-draft/presets/sample-data-b2b/lifttech-solutions-ltd-de-address.spec.ts
```diff
@@ -64,7 +64,7 @@ describe(`with lifttechSolutionsLtdDeAddress preset`, () => {
         "postalCode": "12345",
         "region": "Brandenburg",
         "salutation": undefined,
-        "state": undefined,
+        "state": "Berlin",
```
ByronDWall, line 67: I vaguely remember there being some sort of validation error around having `Berlin` as a state, but it was a while ago and we can always remove this if so

NickDevG: I think that is a problem when the project settings `cart.countryTaxRateFallbackEnabled` is not enabled. But I see is being set `true`

PR #552, opened 2024-04-17
PRC-3436: Add `HighPrecisionMoney` and `HighPrecisionMoneyDraft` models

ragafus:
In Priceless, we want to test Standalone Prices with high precision values. With that goal, and assuming that high precision money could be used elsewhere, this PR introduces `HighPrecisionMoney` and `HighPrecisionMoneyDraft` models to `@commercetools-test-data/commons`.

### DIFF
--- a/models/commons/src/high-precision-money/generator.ts
+++ b/models/commons/src/high-precision-money/generator.ts
@@ -1,0 +1,16 @@
+import { fake, Generator, oneOf } from '@commercetools-test-data/core';
+import { THighPrecisionMoney } from './types';
+
+// https://docs.commercetools.com/api/types#highprecisionmoney
+
+const generator = Generator<THighPrecisionMoney>({
+  fields: {
+    type: 'highPrecision',
+    centAmount: fake((f) => f.number.int({ min: 1, max: 999999 })),
+    currencyCode: oneOf('EUR', 'USD'),
+    fractionDigits: fake((f) => f.number.int({ min: 3, max: 10 })),
+    preciseAmount: fake((f) => f.number.int({ min: 1, max: 9999999999 })),
+  },
+});
+
+export default generator;
--- a/models/commons/src/high-precision-money/generator.ts
+++ b/models/commons/src/high-precision-money/generator.ts
@@ -3,13 +3,22 @@ import { THighPrecisionMoney } from './types';
 
 // https://docs.commercetools.com/api/types#highprecisionmoney
 
+const CENT_PRECISION_FRACTION_DIGITS = 2;
+
 const generator = Generator<THighPrecisionMoney>({
   fields: {
     type: 'highPrecision',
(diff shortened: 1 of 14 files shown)


models/commons/src/high-precision-money/high-precision-money-draft/presets/index.ts
```diff
@@ -0,0 +1,3 @@
+const presets = {};
+
+export default presets;
```
ARRIOLALEO, line 3: there is a need of this ? we are not using the presets in this money type ? .

ARRIOLALEO (approved):

models/commons/src/high-precision-money/generator.ts
```diff
@@ -0,0 +1,16 @@
+import { fake, Generator, oneOf } from '@commercetools-test-data/core';
+import { THighPrecisionMoney } from './types';
+
+// https://docs.commercetools.com/api/types#highprecisionmoney
+
+const generator = Generator<THighPrecisionMoney>({
+  fields: {
+    type: 'highPrecision',
+    centAmount: fake((f) => f.number.int({ min: 1, max: 999999 })),
```
dogayuksel, line 9: I think it would be more coherent if we were to compute centAmount from the generated preciseAmount and fractionDigits. ['postBuild' on the generator](https://github.com/jackfranklin/test-data-bot/blob/1dd9bcc0c10a531f9f7379a7e70bb4c318610d26/README.md?plain=1#L280) should allow that.

models/commons/src/high-precision-money/high-precision-money-draft/generator.ts
```diff
@@ -0,0 +1,16 @@
+import { fake, Generator, oneOf } from '@commercetools-test-data/core';
+import { THighPrecisionMoneyDraft } from '../types';
+
+// https://docs.commercetools.com/api/types#highprecisionmoneydraft
+
+const generator = Generator<THighPrecisionMoneyDraft>({
+  fields: {
+    type: 'highPrecision',
+    centAmount: fake((f) => f.number.int({ min: 1, max: 999999 })),
```
dogayuksel, line 9: Similar situation here, we should either set it to null or compute a valid rounded version. Or it will throw an error when we submit as is:

> A Price of 1.015 USD can be rounded either to 1.01 USD or 1.02 USD. If it lies outside of this range, an error message stating that centAmount must be rounded correctly will be returned. 

https://docs.commercetools.com/api/types#highprecisionmoneydraft

models/commons/src/high-precision-money/high-precision-money-draft/presets/index.ts
ragafus: > there is a need of this ? we are not using the presets in this money type ? .

The presets object is also present in the rest of the models that don't provide any preset. I have no info on why this is the current approach of the package, I assume it is because of consistency for the user of the models.

dogayuksel (approved):

PR #553, opened 2024-04-18
PANGOLIN-3789: updated `code` property on discount-code-draft to fix duplicate value

kterry1:
The `code` property on `discount-code-draft` is causing flaky tests. This PR adds randomness to the `code` property with an alphanumeric value so we can lessen the chance of duplicate value errors.

Example: 
```
Error in function getDiscountCodeCreateData --- Error: 
BadRequest: A duplicate value '"addo"' exists for field 'code'.
```

Reference: https://app.circleci.com/pipelines/github/commercetools/audit-log/26918/workflows/caacfa3c-1f72-4ffd-82e7-3b3b839560fc/jobs/272586/tests

### DIFF
--- a/models/discount-code/src/discount-code-draft/generator.ts
+++ b/models/discount-code/src/discount-code-draft/generator.ts
@@ -12,7 +12,7 @@ const generator = Generator<TDiscountCodeDraft>({
   fields: {
     name: fake(() => LocalizedStringDraft.random()),
     description: fake(() => LocalizedStringDraft.random()),
-    code: fake((f) => f.lorem.word()),
+    code: fake((f) => f.lorem.words({ min: 1, max: 3 })),
     cartDiscounts: [],
     cartPredicate: '1=1',
     isActive: fake((f) => f.datatype.boolean()),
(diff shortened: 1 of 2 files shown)


models/discount-code/src/discount-code-draft/generator.ts
```diff
@@ -12,7 +12,7 @@ const generator = Generator<TDiscountCodeDraft>({
   fields: {
     name: fake(() => LocalizedStringDraft.random()),
     description: fake(() => LocalizedStringDraft.random()),
-    code: fake((f) => f.lorem.word()),
```
ragafus, line 15: WDYT of instead of using `f.lorem.words({ min: 1, max: 3 })` we use `f.string.alphanumeric({ length: 20 })` to reduce collision probability even more?.

ragafus (approved):

models/discount-code/src/discount-code-draft/generator.ts
nima-ct: or increasing the minimum to at least 2?

kterry1: I do like the alphanumeric idea. I'll make that change now. Thanks!

nima-ct (approved):

PR #556, opened 2024-04-23
PRC-3447: Update StandalonePriceDraft model

nima-ct:
https://commercetools.atlassian.net/browse/PRC-3447

This PR updates the StandalonePriceDraft and PriceTierDraft's model to cover realistic scenarios and avoid custom overrides in our tests.

Notes:
- Address the issue with the minimum and maximum quantity generated on price tiers
- Main price, staged price, and tier prices should ideally share the same currency.
The preset solves the main price-tier prices relationships until we find a default approach.
- `tiers` have been set to null to reflect our API constraints regarding non-shared currencies


dogayuksel:
So the currencies of tier prices are not yet generated on the [postBuild hook]( https://github.com/jackfranklin/test-data-bot/blob/1dd9bcc0c10a531f9f7379a7e70bb4c318610d26/README.md?plain=1#L280) of the standalone price generator? This approach forces us to always use the preset to get data that is submittable.

nima-ct:
> So the currencies of tier prices are not yet generated on the [postBuild hook](https://github.com/jackfranklin/test-data-bot/blob/1dd9bcc0c10a531f9f7379a7e70bb4c318610d26/README.md?plain=1#L280) of the standalone price generator? This approach forces us to always use the preset to get data that is submittable.

Ok, I think I misunderstood the `postBuild hook` thing when you mentioned it.
Sounds promising, let me give it a try then.

dogayuksel (approved):

models/standalone-price/src/standalone-price-draft/generator.ts
```diff
@@ -21,7 +20,7 @@ const generator = Generator<TStandalonePriceDraft>({
     channel: fake(() => KeyReferenceDraft.presets.channel()),
     validFrom: fake(getValidFrom),
     validUntil: fake(getValidUntil),
-    tiers: [fake(() => PriceTierDraft.random())],
+    tiers: null,
```
ragafus, line 23: Why don't we keep the current behavior? We might be breaking existing tests due to this change.

models/standalone-price/src/standalone-price-draft/presets/withTiers.ts
```diff
@@ -0,0 +1,21 @@
+import { MoneyDraft, PriceTierDraft } from '@commercetools-test-data/commons';
+import type { TStandalonePriceDraftBuilder } from '../../types';
+import StandalonePriceDraft from '../builder';
+
+const withTiers = (currencyCode: string): TStandalonePriceDraftBuilder => {
```
ragafus, line 5: Should the staged value have the same currency too? Maybe @ARRIOLALEO and @dogayuksel can throw light on this question.

ragafus (commented):
We need to run `pnpm changeset` to bump packages versions.

models/standalone-price/src/standalone-price-draft/generator.ts
dogayuksel: currencies in the tiers has to match main price to fit the spec. Either we mutate it on the transform phase, or like so is going to be an invalid price 50% of the time. We could bump a major version, but doubt anyone else is using this entity. B2B data set is not affected by this.

models/standalone-price/src/standalone-price-draft/presets/withTiers.ts
nima-ct: It is possible to have a different currency for the staged price, it doesn't contradict the specs.
I like the approach of not having to explicitly name it as a unified currency thing and instead having the API spec as a guideline. If it's `withTiers` then the tiers should be valid.

I see the benefit of having a homogeneous currency but I think of it as an additional preset when we'll need for one.

ARRIOLALEO: AFAIK we will not allow the customer to change the currency and the most logic will be that a price contains only one currency, if they need staged price to have another currency means it is in another country, the logic will be that the customer create a new price, in my opinion we have too keep the logic one price one type of currency.

ragafus (approved):

ARRIOLALEO (approved):
thanks for all changes .🚀🚀🚀🚀🚀

.changeset/sharp-kiwis-eat.md
```diff
@@ -0,0 +1,9 @@
+---
+'@commercetools-test-data/standalone-price': major
```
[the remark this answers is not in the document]
nima-ct: @tdeekens
Thanks for pointing it out. I must have followed the prompts too fast after a `pnpm changeset` and thought it was set as major automatically.
There was a slight uncertainty that it could affect other tests (https://github.com/commercetools/test-data/pull/556#discussion_r1578256475) but considering there are no potential breaking changes it's more of a minor bump than a major

PR #559, opened 2024-04-30
PANGOLIN-8320: reverted presets payment draft

kterry1:
Reverting presets for payment draft that were created because of an error in Audit Log now that the error has been handled by the Pangolin Team.


models/commons/src/money/money-draft/generator.ts
```diff
@@ -5,7 +5,7 @@ import { TMoneyDraft } from '../types';
 
 const generator = Generator<TMoneyDraft>({
   fields: {
-    centAmount: fake((f) => f.number.int({ min: 10 })),
+    centAmount: fake((f) => f.number.int({ min: 10, max: 10000 })),
```
kterry1, line 8: To stay consistent with mc-fe.

Sarah4VT (approved):

PR #566, opened 2024-05-06
[FCT-812] Sample Data - BoGo Discount to be Applied Multiple Times

jocelyn-dunkley:
Remove app limit and increase customer limit to 5 for BOGO discounts


jaikamat (approved):
Looks like this change does the thing!

Tests will probably have to be changed as well, just a heads up

stephsprinkle:
We need a changeset!!

tylermorrisford (approved):

PR #577, opened 2024-05-09
chore: return required value type model from cart discount

ARRIOLALEO:
when we return the values from the card cart discount we need a complete price value not a draft, that does not mach with what the api return https://docs.commercetools.com/api/projects/cartDiscounts#cartdiscountvalue draft miss the type property.(draft are not a preset they are a different type)

we are the only team using the card discount which is the reason for omitting the canary realize.

### DIFF
--- a/.changeset/metal-experts-reflect.md
+++ b/.changeset/metal-experts-reflect.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-test-data/cart-discount': patch
+---
+
+return a full price on value instead a draft price to mach the values expected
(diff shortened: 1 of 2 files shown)


.changeset/shiny-singers-nail.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-test-data/cart-discount': minor
```
ragafus, line 2: Nitpick: as we are fixing a wrong behavior, I'd expect changes type to be `patch`

ragafus (approved):

PR #579, opened 2024-05-10
chore: rename ref in graphql

ARRIOLALEO:
This pr Changes and now we add the missing references to the references

this pr will solve the issue the card discount value when giftline item is selected, I will go with a canary Realize becuase we are modify some hard code data on data examples
```      

 Missing field 'distributionChannelRef' while writing result {
      "type": "giftLineItem",
      "product": {
        "id": "07dff654-196d-4ba0-aae6-3508dfaf7d80",
        "typeId": "product",
        "__typename": "Reference"
      },
      "variantId": 341916377939968,
      "supplyChannel": {
        "id": "ba4652d3-5aef-448c-903c-c8abe4bb48cf",
        "typeId": "channel",
        "__typename": "Reference"
      },
      "distributionChannel": {
        "id": "bd072d11-791a-4d67-a34a-40786c72b9df",
        "typeId": "channel",
        "__typename": "Reference"
      },
      "__typename": "GiftLineItemValue"
    }```

this bug what found during https://app.circleci.com/pipelines/github/commercetools/merchant-center-frontend/85029/workflows/7e3521d2-06f8-46e6-bf4b-d2f3fabdf2af/jobs/1888955

when the value on a card discount is created there are 3 options and the tests fails https://github.com/commercetools/test-data/blob/main/models/cart-discount/src/cart-discount/generator.ts#L27

when CartDiscountValueGiftLineItem.random(), is randomly selected

### DIFF
--- a/models/cart-discount/src/cart-discount-value-gift-line-item/builder.spec.ts
+++ b/models/cart-discount/src/cart-discount-value-gift-line-item/builder.spec.ts
@@ -74,15 +74,7 @@ describe('builder', () => {
           typeId: 'product',
         }),
         variantId: expect.any(Number),
-        supplyChannel: expect.objectContaining({
-          id: expect.any(String),
-          typeId: 'channel',
-        }),
         supplyChannelRef: expect.objectContaining({
-          id: expect.any(String),
-          typeId: 'channel',
-        }),
-        distributionChannel: expect.objectContaining({
           id: expect.any(String),
           typeId: 'channel',
         }),
--- a/models/cart-discount/src/cart-discount-value-gift-line-item/builder.spec.ts
+++ b/models/cart-discount/src/cart-discount-value-gift-line-item/builder.spec.ts
@@ -22,11 +22,11 @@ describe('builder', () => {
           typeId: 'product',
         }),
         variantId: expect.any(Number),
-        supplyChannelRef: expect.objectContaining({
+        supplyChannel: expect.objectContaining({
           id: expect.any(String),
           typeId: 'channel',
         }),
-        distributionChannelRef: expect.objectContaining({
+        distributionChannel: expect.objectContaining({
           id: expect.any(String),
           typeId: 'channel',
         }),
@@ -48,11 +48,11 @@ describe('builder', () => {
           typeId: 'product',
         }),
         variantId: expect.any(Number),
-        supplyChannelRef: expect.objectContaining({
+        supplyChannel: expect.objectContaining({
           id: expect.any(String),
           typeId: 'channel',
         }),
-        distributionChannelRef: expect.objectContaining({
+        distributionChannel: expect.objectContaining({
           id: expect.any(String),
           typeId: 'channel',
         }),
@@ -74,7 +74,15 @@ describe('builder', () => {
           typeId: 'product',
         }),
         variantId: expect.any(Number),
+        supplyChannel: expect.objectContaining({
+          id: expect.any(String),
+          typeId: 'channel',
+        }),
         supplyChannelRef: expect.objectContaining({
+          id: expect.any(String),
+          typeId: 'channel',
+        }),
+        distributionChannel: expect.objectContaining({
           id: expect.any(String),
           typeId: 'channel',
         }),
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by dogayuksel** [comment]
@ARRIOLALEO Are you sure these `...ref` properties really belong to the draft entity? When checking generated type declarations, I don't see these properties. I only see `...ref` ones on the entities we fetch. This might be a quirk of the discount application, for which we do custom transformations [here](https://github.com/commercetools/merchant-center-frontend/blob/main/packages-application/application-discounts/src/utils/graphql-tools/transformers.js#L184). Maybe these changes don't belong to the test-data, it seems like it was already implemented according to the spec.

File: models/cart-discount/src/cart-discount-value-gift-line-item/cart-discount-value-gift-line-item-draft/builder.spec.ts [hunk 1]
```diff
@@ -74,11 +74,11 @@ describe('builder', () => {
             typeId: 'product',
           }),
           variantId: expect.any(Number),
-          supplyChannel: expect.objectContaining({
+          supplyChannelRef: expect.objectContaining({
```
tdeekens, line 77: Refs usually do not exist in GraphQL. They're a REST API concept. I am not sure we should be adding them to this model as a result.

tdeekens: @dogayuksel had a look at the GraphQL introspection schema

CleanShot 2024-05-10 at 11 45 58@2x

and then GraphiQL. I think these refs exist on the value of cart discounts for instance

CleanShot 2024-05-10 at 11 49 24@2x

ARRIOLALEO: yes i understand that part but i have this shape on card discounts, without this change our tests gets flaky when using the card discount model and by luck we get one gift line items.
``` 

      supplyChannelRef {
        id
        typeId
        __typename
      }
      distributionChannelRef {
        id
        typeId
        __typename
      }
    }
```

 https://github.com/commercetools/merchant-center-frontend/blob/main/packages-application/application-discounts/src/components/cart-discounts/details/cart-discount-details-connector/cart-discount-details-connector.ctp.graphql#L52-L62

ARRIOLALEO: but as @doga infers i will close this pr and investigate further. thanks both your your inputs

ARRIOLALEO: @dogayuksel  @tdeekens  how Is recomended to proceed here, my idea was to pass this info to BE and check the docs.  i can by pass this on our app adding a value when creating a card discount but i think we lost the idea here about having random value.

Im open to recomendations.

dogayuksel: `...ref` versions should be added to the non-draft side. Those seem to be implemented off from the spec, at least that is what is see looking at the generated types.

tdeekens: I would argue they only exist on the none draft type first of all. Moreover, we can then fix that here our pass values when we use the model to restrict randomness. I'd like to learn why the actually exist if we add them here. Maybe a BE engineer can help.

ARRIOLALEO: I already asked our BE I will post here when I get more info, thanks for your help.

ARRIOLALEO: what stops me from doing that is I will need to modify the types that come from the platform, https://github.com/commercetools/commercetools-sdk-typescript/blob/ae315373964a2daba2186a38b795f708b4dc4058/packages/platform-sdk/src/generated/models/cart-discount.ts#L481

we use them to generate, i can omit some fields and add the other but I want to confirm with BE if rest and graphql
 bring the same fields if they are different we will fix one part and break another.

ARRIOLALEO:
@Doga the changes are ready thanks for all your inputs you were right after thinking i think this will make the trick

dogayuksel (approved):

nima-ct (approved):

PR #580, opened 2024-05-10
Ma add missing building fields to models cart discount anddiscount code

ARRIOLALEO:


dogayuksel:
shouldn't it be added to default and rest builds too?

ARRIOLALEO:
added and in the discount code they already exist https://github.com/commercetools/test-data/blob/main/models/discount-code/src/discount-code-draft/transformers.ts#L14

ragafus (approved):

PR #590, opened 2024-05-16
Stores can not be undefined

ARRIOLALEO:
initialize stores


ARRIOLALEO:
not necesary tested on local all working properly

PR #591, opened 2024-05-17
refactor: use generate types for custom view

Rhotimee:
#### Summary

- Use generate types for custom view
- Include custom application installation in customView

### DIFF
--- a/.changeset/lovely-waves-notice.md
+++ b/.changeset/lovely-waves-notice.md
@@ -1,0 +1,6 @@
+---
+'@commercetools-test-data/custom-view': patch
+---
+
+- Use generate types for custom view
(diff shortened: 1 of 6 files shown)


.changeset/lovely-waves-notice.md
```diff
@@ -0,0 +1,6 @@
+---
+'@commercetools-test-data/custom-view': patch
+---
+
+- Use generate types for custom view
+- Include custom application installation in custom view
```
kark, line 6:
```suggestion
- Use generated types for custom view
- Include custom view installation in custom view
```
Maybe we should simply keep the first line

kark (approved):
Thanks!

PR #602, opened 2024-06-17
feat: build `applicationCount` property of GraphQL `DiscountCode` type

ragafus:
This PR builds by default the `applicationCount` prop of DiscountCode.
It's a prop that Priceless is already using in Discounts app.

### DIFF
--- a/models/discount-code/src/transformers.ts
+++ b/models/discount-code/src/transformers.ts
@@ -1,5 +1,6 @@
 import { LocalizedString } from '@commercetools-test-data/commons';
 import { Transformer } from '@commercetools-test-data/core';
+import { faker } from '@faker-js/faker';
 import type { TDiscountCode, TDiscountCodeGraphql } from './types';
 
 const transformers = {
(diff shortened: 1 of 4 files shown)


mario-priceless (approved):

models/discount-code/src/transformers.ts
```diff
@@ -30,6 +30,7 @@ const transformers = {
       'cartDiscounts',
     ],
     addFields: ({ fields }) => ({
+      applicationCount: fields.applicationVersion,
```
dogayuksel, line 33: `applicationVersion` = `applicationCount` ?😕

I would expect this to be on the [generator](https://github.com/commercetools/test-data/blob/f7fe5b7f9dedfca6a2612ce09e515a0125572074/models/discount-code/src/generator.ts#L15), but can't see the prop on the [docs](https://docs.commercetools.com/api/projects/discountCodes#discountcode) either.
Is this going to be an only graphql field?

ragafus: > `applicationVersion` = `applicationCount` ?😕
> 
> I would expect this to be on the [generator](https://github.com/commercetools/test-data/blob/f7fe5b7f9dedfca6a2612ce09e515a0125572074/models/discount-code/src/generator.ts#L15), but can't see the prop on the [docs](https://docs.commercetools.com/api/projects/discountCodes#discountcode) either. Is this going to be an only graphql field?

I only found the property in GraphQL docs. 
I have no idea if `applicationVersion` = `applicationCount`, so, I've added a commit where `applicationCount` is randomly generated.

dogayuksel (approved):

PR #611, opened 2024-07-02
fix: add missing `type`, `fractionDigits` fields to `Money` GraphQL model

ragafus:


ragafus:
Useless: `CentPrecisionMoney` and `HighPrecisionMoney` are already available.

PR #625, opened 2024-08-07
fix(attribute-enum-type): add missing results wrapper

dogayuksel:
- Wraps PlainEnumValue type with results, as done on `attribute-localized-enum-type` as seen below:
https://github.com/commercetools/test-data/blob/827861fe7d04e297cfed09efd4646659f8e260d1/models/product-type/src/attribute-localized-enum-type/transformers.ts#L22-L31
- Tested the updated `@commercetools-test-data/product-type` package locally against all products and discounts tests where it is being used.


CarlosCortizasCT (approved):
Looks good to me 👍 

Thanks!

PR #626, opened 2024-08-07
fix(attribute-enum-type): add changeset

dogayuksel:
Adds changeset for https://github.com/commercetools/test-data/pull/625 😞


CarlosCortizasCT (approved):
Thanks! 🙇

PR #627, opened 2024-08-08
Fix some minor issues with different models

CarlosCortizasCT:
## Description

While working in the MC frontend `rosie` migration ([reference](https://commercetools.atlassian.net/browse/FEC-17)) we noticed some issues with the test data models used in the products tests.

Here we try to address the ones found so far.

Please see inline comments for more details.


models/category/src/transformers.ts
```diff
@@ -2,52 +2,30 @@ import { LocalizedString } from '@commercetools-test-data/commons';
 import { Transformer } from '@commercetools-test-data/core';
 import type { TCategory, TCategoryGraphql } from './types';
 
+const buildFields: Array<keyof TCategory> = [
```
CarlosCortizasCT, line 5: Just a little refactor since all transformers build the same fields.

models/product/src/product-data/transformers.ts
```diff
@@ -59,7 +59,7 @@ const transformers = {
     },
   }),
   graphql: Transformer<TProductData, TProductDataGraphql>('graphql', {
-    buildFields: ['categories'],
+    buildFields: ['categories', 'masterVariant'],
```
CarlosCortizasCT, line 62: We need the graphql transformer to build the `masterVariant` because otherwise consuming the generated data will fail.

models/product/src/product-variant/presets/boring-generic-milk-master-variant.ts
```diff
@@ -1,16 +1,12 @@
...
   ProductVariant()
     .key('boring-generic-milk-master-variant-key')
     .sku('boring-generic-milk-master-variant-sku')
-    .attributes([
-      {
-        name: 'cow-name',
-        value: 'unknown',
-      },
-    ])
+    .attributes([Attribute.random().name('cow-name').value('unknown')])
```
CarlosCortizasCT, line 9: This is a buildable field so we can't provide a value object but use the appropriate test model.

models/product/src/product-variant/presets/happy-cow-milk-master-variant.ts
```diff
@@ -6,14 +7,8 @@ const happyCowMilkMasterVariant = (): TProductVariantBuilder =>
...
     .attributes([
-      {
-        name: 'cow-name',
-        value: 'Buryonka',
-      },
-      {
-        name: 'lactose-free',
-        value: false,
-      },
+      Attribute.random().name('cow-name').value('Buryonka'),
```
CarlosCortizasCT, line 10: Same as above.

models/store/src/store/generator.ts
```diff
@@ -26,7 +26,7 @@ const generator = Generator<TStore>({
     key: fake((f) => f.lorem.slug()),
     name: fake(() => LocalizedString.random()),
     languages: [oneOf('en-US', 'de-DE', 'es-ES')],
-    countries: [fake((f) => f.location.countryCode())],
+    countries: [fake((f) => ({ code: f.location.countryCode() }))],
```
CarlosCortizasCT, line 29: The actual shape of this property is not an array of strings but an array of objects containing a `code` property with the ISO country code.

As a follow up, it would be great if we could make typescript to warn us when we try to provide an invalid value to a property of the model at development time.
We already have [this proposal](https://commercetools.atlassian.net/browse/FEC-62) for the fe chapter.

models/store/src/store/transformers.ts
```diff
@@ -2,25 +2,21 @@ import { LocalizedString, Reference } from '@commercetools-test-data/commons';
 import { Transformer } from '@commercetools-test-data/core';
 import type { TStore, TStoreGraphql } from './types';
 
+const buildFields: Array<keyof TStore> = [
```
CarlosCortizasCT, line 5: Just a little refactor since all transformers build the same fields.

models/product/src/product-data/types.ts
```diff
@@ -13,7 +13,7 @@ import { ValueOf } from '@commercetools-test-data/core/src/@jackfranklin/test-da
 // The base generator model. Consumers configure these fields.
 export type TProductData = Omit<ProductData, 'categories'> & {
   categories: Array<Category>;
-  searchKeyword?: Array<SearchKeyword> | null;
+  searchKeyword?: Array<SearchKeyword>;
```
dogayuksel, line 16: isn't it supposed to be `searchKeywords` without `?`, can't see `searchKeyword` as singular on docs 😕 
https://docs.commercetools.com/api/projects/products#productdata
```suggestion
  searchKeywords: Array<SearchKeyword>;
```

dogayuksel (approved):
We should add the changesets

models/product/src/product-data/types.ts
CarlosCortizasCT: That's a good point.
I was puzzled about having both `searchKeyword` and `searchKeywords` in the [generator](https://github.com/commercetools/test-data/blob/main/models/product/src/product-data/generator.ts#L31). Perhaps it was a mistake when they were introduced or we are missing something.
I know it was a long time ago but maybe @jaikamat can help us here.

Anyway, I don't feel comfortable renaming the field right away because, technically, that would be a breaking change.
What I've done is to add the `searchKeywords` (plural) attribute to the type.

CarlosCortizasCT:
> We should add the changesets

Thanks for the reminder 🙇 

Added them here: ddb43c700d076b62a77ef9c226582d05d9d6af77

models/product/src/product-data/types.ts
jaikamat: image
 
 `searchKeyword` is used in the GraphQL transformation of this model

jaikamat: >I was puzzled about having both searchKeyword and searchKeywords in the [generator](https://github.com/commercetools/test-data/blob/main/models/product/src/product-data/generator.ts#L31).

If GraphQL and REST types sufficiently diverge, we instead use the base generator to aggregate all possible fields, often with the widest type, so that consumers can pass what they need, and then use the transform target (REST or GraphQL) to pick what is needed

models/product/src/product-data/types.ts
```diff
@@ -13,6 +13,7 @@ import { ValueOf } from '@commercetools-test-data/core/src/@jackfranklin/test-da
 // The base generator model. Consumers configure these fields.
 export type TProductData = Omit<ProductData, 'categories'> & {
   categories: Array<Category>;
+  searchKeywords?: Array<SearchKeyword>;
```
jaikamat, line 16: We shouldn't need to include this as it's already encapsulated in the `Omit<ProductData, 'categories'>` type

bradd123 (approved):
💯 Thanks for the changes

models/product/src/product-data/transformers.ts
jaikamat: Nice catch 👍

models/product/src/product-data/types.ts
CarlosCortizasCT: Thanks Jai 🙇 

Removed here: dfc263c657ba6c4f9942f98fcc24e83379f9d381


Just out of curiosity. What do we need the `searchKeyword` property for?

jaikamat: See [this comment](https://github.com/commercetools/test-data/pull/627#discussion_r1709482883), not sure if it got missed in resolution

jaikamat (approved):
Thank you for tackling this! 🏅

PR #629, opened 2024-08-09
Update management of locale-based props for GraphQL transformers

CarlosCortizasCT:
## Description

There are some properties in the test data models that are locale based but we're not handling them in a way that resembles the way they are handled in the GraphQL context.

Let's take the Category model `name` property as an example.

In the Core definition for that model, the `name` property is a `LocalizedString` type, which is an object that contains a map of locale to string, so that's what we configure [in the model's generator](https://github.com/commercetools/test-data/blob/main/models/category/src/generator.ts#L20).

However, the GraphQL representation of that property is defined as an array of [TLocalizedStringGraphql](https://github.com/commercetools/test-data/blob/main/models/commons/src/localized-string/types.ts#L6) type, which is an object that contains the locale and the value.

Also, in GraphQL we prefer to use the calculated **nameAllLocales** property which [is created](https://github.com/commercetools/test-data/blob/main/models/category/src/transformers.ts#L30) in the GraphQL transformer of the Category model, but that does not mean that the **name** property can't be used. Actually, in the MC frontend we are using it ([example](https://github.com/commercetools/merchant-center-frontend/blob/main/packages-application/application-products/src/components/pim-search-list/pim-product-list-connector/products.ctp.graphql#L109)).

The issue is that, if you check the previous example, the expected **name** property is just a string, but the data model is returning a `TLocalizedStringGraphql` object instead.
This is how the Category entity looks like in the GraphQL API:
image

The GraphQL api is expecting the **name** to be consumed using a locale so that the server will return a string in the requested locale.

So, the GraphQL transformer for the Category model should be updated to handle the **name** property in a way that it returns a string instead of an array of `TLocalizedStringGraphql` objects.

I've been thinking about different approaches and the one I'm proposing here is about the GraphQL transformer to not build the **name** property automatically, "manually" build the `nameAllLocales` property and then use the latter to get the value we will be replacing the **name** property for so instead of returning a `LocalizedString` object, we will be returning a string.

Since the `LocalizedString` value of the `name` property which gets to the GraphQL transformer will contain potentially several names for different locales, we would be returning the english (decided default locale) version, or the first one in the list if the english version is not present.
(_Using english as the default locale is what we were [already doing](https://github.com/commercetools/merchant-center-frontend/blob/main/packages-shared/test-data/src/product-data/transformers.js#L78) in the local test-data of the MC repository_)

This way we can keep a very similar API from the consumer perspective (they can configure different names per locale which can be consumed in the `nameAllLocales` property) and know the plain string `name` value will be the english version they provided. 

There's an alternative approach I'm exploring here: #630

### DIFF
--- a/.changeset/clever-taxis-carry.md
+++ b/.changeset/clever-taxis-carry.md
@@ -1,0 +1,24 @@
+---
+'@commercetools-test-data/category': patch
+---
+
+Fixed the return type for some properties in the GraphQL representation:
+
+- name
+- description
+- slug
+- metaTitle
+- metaKeywords
+- metaDescription
+
+They all were returning a value with type `TLocalizedStringGraphql` where they should be returning a `string`.
+
+From the consumer point of view, mocking those values won't change so they can still use the `LocalizedString` common model to set it up:
+
+```js
+Category.random()
+  .name(LocalizedString.random().en('Shoes').de('Schuhe'))
+  .buildGraphql();
+```
+
+The actual `string` returned will be the one with the default locale (`en`) or the first one if that one does not exist.
--- a/.changeset/curvy-cycles-switch.md
+++ b/.changeset/curvy-cycles-switch.md
@@ -1,0 +1,24 @@
+---
+'@commercetools-test-data/product': patch
+---
+
+Fixed the return type for some properties in the GraphQL representation:
+
+- name
+- description
+- slug
+- metaTitle
+- metaKeywords
+- metaDescription
+
+They all were returning a value with type `TLocalizedStringGraphql` where they should be returning a `string`.
+
+From the consumer point of view, mocking those values won't change so they can still use the `LocalizedString` common model to set it up:
+
+```js
+Product.random()
+  .name(LocalizedString.random().en('Shoes').de('Schuhe'))
+  .buildGraphql();
+```
+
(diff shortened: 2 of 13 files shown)


ahmehri (approved):

tdeekens (approved):
Thanks for the elaborate description. This really helped me and I read it 2-3 times. 

For context: the name notation returning a string was first in the GraphQL APIs. It serves more the shop-oriented use case. The nameAllLocales was added after actually for the MC as it gives more flexibility for inputs expanding into various languages. Maybe this was obvious but wanted to share regardless. 

I've been flipping between PRs for some time now. Having a hard time deciding. To me, to the outside, both are similar and fix the issue. They're different mainly internally how they get there. 

While I find the second slightly more elegant, I am unsure if the typing issue is a hill worth dying on. Given that this approach seems very reasonable. 

I also toyed alternative approach in which I wanted to use the ability to use a function as the second argument of a builder's setter (in code fnOrValue). To there take a locale from the object of locales. That however would break transforming as you'd wipe the models state for other locales. 

I think this approach is sane and I also can not think of a better alternative.

emmenko (approved):
Thanks a lot for finding some solutions for this edge case. Definitely not an easy one.

I think I also prefer this approach, I think it should be good enough to cover this edge case without disrupting much the rest of the setup.

All good then 🤗

dogayuksel (approved):

PR #630, opened 2024-08-12
Update management of locale-based props for GraphQL transformers (alternative)

CarlosCortizasCT:
## Description

This is an alternative proposal for the issue explained in #629.

In this one, we let the GraphQL transformer build the **name** property (to a [TLocalizedStringGraphql](https://github.com/commercetools/test-data/blob/main/models/commons/src/localized-string/types.ts#L6) type), we use it directly to return the `nameAllLocales` property value, and we replace the `name` property with the default locale of the latter.

The issue here is the typing for the `replaceFields` transformer fuction.
Currently the `fields` property that function is called with, is defined as the original model (`TCategory` in this case) but that's not actually true as the "buildable" fields have already been built with the respective graphql transformers. However, we can't just change the type of the `fields` property to `TCategoryGraphql` because over there we need the name to be a `string` but the value we get from the build phase of that property is a `TLocalizedStringGraphql` type.

I was trying to make the `replaceFields` function more generic so that we could provide a new type describing the shape of the actual `fields` property that the function is called with. However, this is not working for me yet (typescript issues) and maybe it's not even the right approach.

### DIFF
--- a/core/src/transformer.ts
+++ b/core/src/transformer.ts
@@ -58,7 +58,7 @@ function Transformer<Model, TransformedModel>(
         );
       }
       return fieldsReplacer({
-        fields: transformedFields,
+        fields: transformedFields as unknown as TransformedModel,
       }) as unknown as TransformedModel;
     }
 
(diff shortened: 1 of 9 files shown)


CarlosCortizasCT:
Closing in favor of #629

PR #634, opened 2024-08-16
FEC-6: Add missing build fields to customer and order models

dogayuksel:
- Add `customer` property to Order model build fields on default and graphql transforms.
- Fix Customer model to build `addresses` and `stores` on all transforms.
- Allow Customer model to receive builders for `defaultShippingAddress`, `shippingAddresses`, `defaultBillingAddress`, and `billingAddresses` properties. Adds these properties to build fields for default and graphql transforms.


models/customer/src/generator.ts
```diff
@@ -32,11 +32,15 @@ const generator = Generator<TCustomer>({
     dateOfBirth: null,
     companyName: fake((f) => f.company.name()),
     vatId: fake((f) => f.string.alphanumeric(12)),
-    addresses: fake(() => Address.random()),
+    addresses: fake(() => [Address.random()]),
```
antoniolodias, line 35: I don't know exactly how flexible we have to be, but I think addresses can eventually be just an empty array

dogayuksel: should be okay I think, then one could do: `Customer.random().addresses([]).buildGraphql()`

antoniolodias: sounds good!

antoniolodias (approved):

models/customer/src/transformers.ts
```diff
@@ -1,15 +1,36 @@
...
 import type { TCustomer, TCustomerGraphql } from './types';
 
+const commonBuildFields: (keyof TCustomer)[] = [
+  'createdBy',
+  'lastModifiedBy',
+  'stores',
+  'addresses',
+];
+
+const removeFields: (keyof TCustomer)[] = [
```
CarlosCortizasCT, line 11: Since these ones are not shared and we use them only in the `rest` transformer, would it be more easy to follow if we just set them inline?

models/customer/src/types.ts
```diff
@@ -1,7 +1,13 @@
...
 import type { TBuilder } from '@commercetools-test-data/core';
 
-export type TCustomer = Customer;
+export type TCustomer = Customer & {
+  defaultShippingAddress: TAddress | null;
+  shippingAddresses: TAddress[] | null;
+  defaultBillingAddress: TAddress | null;
+  billingAddresses: TAddress[] | null;
+};
 export type TCustomerDraft = CustomerDraft;
```
CarlosCortizasCT, line 11: These new properties that we're adding in the `TCustomer` type will only exist in the GraphQL representation of this model but we need to add them here so consumers can mock them. Is this statement right?

models/customer/src/transformers.ts
dogayuksel: https://github.com/commercetools/test-data/pull/634/commits/2c0abbc1ff15aa5aa7282cee028c4bdbcd8b9339

CarlosCortizasCT (approved):
Many thanks, Doga 🙇

PR #638, opened 2024-08-20
[FEC-38] Fix `ProductProjection` model reference properties

CarlosCortizasCT:
## Description

It all started when migrating [some tests](https://github.com/commercetools/merchant-center-frontend/blob/main/packages-application/application-products/src/product-details/product-details-page.spec.js) in the MC Products application which started to fail when I used the `ProductProjection` model.

The issue was in [this specific function](https://github.com/commercetools/merchant-center-frontend/blob/main/packages-shared/experimental-components/src/utils/products/products.jsx#L79) as we are accessing some properties of the `productType` property of the [rest representation](https://docs.commercetools.com/api/projects/productProjections#productprojection) for the `ProductProjection` model.

In the test data models, whenever we need to deal with a reference, we use the `Reference` model which creates an object with an `id` and `typeId` properties by default. Also, in the rest transformers, we create an `obj` property which is supposed to hold the referenced object model but we're creating that one only with its ID and nothing else. That means that, when tests need to access any property of that referenced model, it's not there.

First thing I did is to update the `Reference` model to allow customization of the `obj` property so consumers can actually populate the reference with an actual referenced object.
Then I had to update the `ProductProjection` model so all references are created with a proper `obj` property.

After that, I also found another models that were not using references correctly so I needed to update them as well.

Technically speaking this could be a breaking change because we're changing the types of some models properties (specially in the REST version) but I see them more like a fix because the types we were using didn't seem correct.

**TODO**
- There's a build error I don't understand. Any help would be appreciated.


models/standalone-price/src/types.ts
```diff
@@ -4,13 +4,18 @@ import {
...
   TReferenceGraphql,
+  TReferenceRest,
 } from '@commercetools-test-data/commons';
 import type { TBuilder } from '@commercetools-test-data/core';
+import { TCustomerGroup } from '@commercetools-test-data/customer-group';
 
 // Base representation
+// TODO: This doesn't seem to be right. I don't understand why the default type should have customized fields.
+// It seems to me that's something only transformed versions of the model should have.
 export type TStandalonePrice = Omit<
```
CarlosCortizasCT, line 19: This is something I don't understand. Maybe it was set up like this to notice these properties have a different type in the alternative representations of the model (rest and graphql) but this pattern is actually missleading.

models/product-projection/src/transformers.ts
```diff
@@ -79,35 +79,69 @@ const transformers = {
...
+          .id(restProductType.id)
+          .key(restProductType.obj?.key || '')
+          .name(restProductType.obj?.name || '')
+          .buildGraphql<TProductTypeGraphql>();
+
+        let state: TStateGraphql | null = null;
+        let stateRef: TReferenceGraphql | null = null;
+        if (fields.state) {
+          const restState = buildField(fields.state, 'rest');
+          stateRef = buildField(fields.state, 'graphql') as TReferenceGraphql;
```
CarlosCortizasCT, line 99: Since the nested model in the default representation holds a `Reference`, it would be great that we could build the `obj` property of the reference for the graphql output but we can't build nested properties in isolation and the graphql transformer of the `Reference` model removes the `obj` property since it's not part of that representation. So the alternative I found is to build the rest representation and use some its values to create a new graphql model and populate it with it.

Technically this is not 100% right as we would need to overwrite all create graphql instance with the values from the rest one but I can't see a way to do it in a maintainable way.

nima-ct:
Hey @CarlosCortizasCT, I just finished a fix on the StandalonePrice model transformers. If that can be of any help, let me know
https://github.com/commercetools/test-data/pull/639

CarlosCortizasCT:
> Hey @CarlosCortizasCT, I just finished a fix on the StandalonePrice model transformers. If that can be of any help, let me know #639

Hi @nima-ct 

Thanks for the heads up.

It seems the work of your PR is part of what I've done here.
I'll take a deeper look to yours in case I've missed something.

ahmehri (approved):

CarlosCortizasCT:
Superseded by #738

PR #639, opened 2024-08-20
Fix GQL export of the StandalonePrice Model

nima-ct:
This update was initiated because we encountered missing field errors during testing with the StandalonePriceModel.
The returned GraphQL data from the mock didn’t match the schema expected by the Apollo client.

This PR addresses issues where the previous logic failed to return correct values for `channelRef` and `customerGroupRef` in the case of a GraphQL build, partially because of a glitch in one of the transformer file functions.

The changes made in the `builder.spec.ts` file ensure that the `id` field is used instead of the `key` field for both `channelRef` and `customerGroupRef`.

With these changes, in the case of a GraphQL build, the `CustomerGroupRef` and `ChannelRef` return ids, not keys, which aligns with our GraphQL schema.


models/standalone-price/src/transformers.ts
```diff
@@ -97,18 +87,27 @@ const transformers = {
       'channel',
       'tiers',
     ],
-    addFields: ({ fields }) => {
+    replaceFields: ({ fields }) => {
```
nima-ct, line 90: `replaceFields` correctly retrieves the `fields` values.
 
When `addFields` is invoked, within the body of that function, every property of the `fields` object is interpreted as an anonymous function, for some reason.

Therefore, the previous implementation of the `addField` function was irrelevant because it relied on precise property values (`fields.customerGroup.id `and `fields.channel.id)`, which were instead evaluated as anonymous functions, causing a mismatch of id's and key's in refs values.

models/standalone-price/src/builder.spec.ts
```diff
@@ -166,6 +167,11 @@ describe('builder', () => {
...
         __typename: 'StandalonePrice',
       })
-    )
-  );
+    );
+
+    expect(standalonePrice.customerGroup?.id).toBe(
+      standalonePrice.customerGroupRef?.id
+    );
+    expect(standalonePrice.channel?.id).toBe(standalonePrice.channelRef?.id);
```
nima-ct, line 175: These values weren’t synchronized before due to a glitch in the transformer’s addFields function. See below

models/standalone-price/src/transformers.ts
```diff
@@ -97,18 +87,27 @@ const transformers = {
...
+            .id(fields.customerGroup.id)
+            .buildGraphql<TReferenceGraphql<'customer-group'>>()
         : null;
+
       const channelRef = fields.channel
-        ? KeyReference.random()
+        ? Reference.random()
             .typeId('channel')
-            .key(fields.channel.key)
+            .id(fields.channel.id)
```
nima-ct, line 101: AFAIK in both the GraphQL schema and the `TStandalonePriceGraphql` within the repository, the `channelRef` and `customerRef` fields return `id` values not `key`s.

jaikamat (approved):
This was an excellent catch, great work! 🏅

models/standalone-price/src/transformers.ts
```diff
@@ -14,19 +14,11 @@ import type {
 
 // Overloads
 function createCustomerGroupReference(
```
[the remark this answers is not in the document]
nima-ct: Good catch.
I wasn't sure tbh but it seems it was a type safety thing because the function was handling 2 different targets and types of output.
Not needed anymore

models/standalone-price/src/transformers.ts
```diff
@@ -14,19 +14,11 @@ import type {
...
-  customerGroup: TCustomerGroup,
-  target: 'graphql'
-): TReferenceGraphql<'customer-group'>;
 
 // Create a customer group reference depending on `id` or `key` field availability
-function createCustomerGroupReference(
-  customerGroup: TCustomerGroup,
-  target: 'graphql' | 'rest'
-) {
+function createCustomerGroupReference(customerGroup: TCustomerGroup) {
```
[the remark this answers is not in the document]
nima-ct: True. I see that `CustomerGroupReference` and `ChannelReference` use IDs in HTTP API spec

models/standalone-price/src/builder.spec.ts
```diff
@@ -108,10 +110,11 @@ describe('builder', () => {
     )
   );
 
-  it(
-    ...createBuilderSpec<TStandalonePrice, TStandalonePriceGraphql>(
-      'graphql',
-      StandalonePrice.random(),
+  it('should build properties for "graphql"', () => {
```
dogayuksel, line 113: lets use the original wrapper to keep things coherent

dogayuksel (approved):

models/standalone-price/src/transformers.ts
ByronDWall: from `packages-shared/types/src/generated/ctp.ts`:

```
export type TChannelReferenceIdentifier = {
  __typename?: 'ChannelReferenceIdentifier';
  id?: Maybe<Scalars['String']['output']>;
  key?: Maybe<Scalars['String']['output']>;
  typeId: Scalars['String']['output'];
};
```

```
export type TCustomerGroupReferenceIdentifier = {
  __typename?: 'CustomerGroupReferenceIdentifier';
  id?: Maybe<Scalars['String']['output']>;
  key?: Maybe<Scalars['String']['output']>;
  typeId: Scalars['String']['output'];
};
```

@nima-ct `key` is a valid field for graphql references of both `customerGroup` and `channel`

ByronDWall: but, youre right, `customerGroupRef` only allows for `id` in the graphql schema

nima-ct: Hi @ByronDWall, I’m not sure if the types you mentioned apply to the StandalonePrice. In the ctp file, both channelRef and customerGroupRef are typed as Reference for their values.
```
export type TStandalonePrice = TVersioned & {
  __typename?: 'StandalonePrice';
  ...
  channelRef?: Maybe<TReference>;
  customerGroupRef?: Maybe<TReference>;
};
```
```
export type TReference = {
  __typename?: 'Reference';
  id: Scalars['String']['output'];
  typeId: Scalars['String']['output'];
};
```

ByronDWall: `TResourceIdentifierInput` allows a key and is used in `Mutation.CreateStandalonePrice`, but not in `Query.StandalonePrice`.

Your PR is correct, and I think this is where the original confusion originated from.

PR #645, opened 2024-08-23
Models refactor proposal

CarlosCortizasCT:
_TBD_


core/src/builder.ts
```diff
@@ -135,6 +136,17 @@ function Builder<Model>({
                   break;
               }
 
+              // TODO: Super hack to try some things out
+              if (type === 'rest') {
```
tdeekens, line 140: This would go away once all builders exist in two variations, right?

core/src/helpers.ts
```diff
@@ -168,6 +172,41 @@ const buildRestList = <Model>(
...
+  type: 'rest' | 'graphql';
+  buildFields?: (keyof TModel)[];
+};
+const createSpecializedBuilder = <TModel>(
+  params: TCreateSpecializedBuilderParams<TModel>
+) => {
+  const modelBuilder = Builder<TModel>({
+    type: params.type,
+    generator: params.generator,
+    transformers: createSpecializedTransformers<TModel>({
```
tdeekens, line 201: Conceptually, thinking about it: we remove the need for transformers and do less heavy lifting in them. In turn we have indirection which needs to be less frequently used/visted.

models/order-alternative/src/builders.ts
```diff
@@ -0,0 +1,54 @@
...
+    generator: graphqlGenerator,
+    type: 'graphql',
+    buildFields: [
+      ...(commonBuildFields as (keyof TOrderGraphql)[]),
+      'customerGroupRef',
+      'businessUnitRef',
+      'storeRef',
+      'cartRef',
+      'quoteRef',
+      'stateRef',
```
tdeekens, line 52: This is clean 💯

models/order-alternative/src/generators.ts
```diff
@@ -0,0 +1,117 @@
...
+  shipping: [],
+  taxMode: oneOf(...Object.values(taxMode)),
+  taxRoundingMode: oneOf(...Object.values(taxRoundingMode)),
+  taxCalculationMode: oneOf(...Object.values(taxCalculationMode)),
+  country: fake((f) => f.location.countryCode()),
+  orderState: oneOf(...Object.values(orderState)),
+  shipmentState: oneOf(...Object.values(shipmentState)),
+  paymentState: oneOf(...Object.values(paymentState)),
+  // shippingInfo: null,
+  syncInfo: null,
```
tdeekens, line 65: I think this was so that they're null in the GraphQL version. Just a guess.

models/order-alternative/src/index.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export * from './types';
+export * as constants from './constants';
+
+export const OrderRest = {
+  random: RestModelBuilder,
+};
+export const OrderGraphql = {
+  random: GraphqlModelBuilder,
+};
```
tdeekens, line 19: Just a thought: we don't need a barrel export file. We could also go `orders/rest`, `orders/graphql` and `orders/draft` and import as `import { Order as OrderGraphQl } from '@commercetools/test-data/order/graphql` for instance.

models/order-alternative/src/order-from-cart-draft/generators.ts
```diff
@@ -0,0 +1,25 @@
...
+    purchaseOrderNumber: fake((f) => String(f.number.int({ min: 100000 }))),
+    paymentState: oneOf(...Object.values(paymentState)),
+    orderState: oneOf(...Object.values(orderState)),
+    state: fake(() => ReferenceDraft.random().typeId('state')),
+    shipmentState: oneOf(...Object.values(shipmentState)),
+    custom: null,
+  },
+});
+
+export const graphqlGenerator = restGenerator;
```
tdeekens, line 25: If drafts are always the same between REST and GraphQL, do we need to export both variations or could we have one?

models/product-projection-alternative/src/presets/happy-cow-milk-product-projection.ts
```diff
@@ -0,0 +1,79 @@
...
+
+    const productDescription = LocalizedString.presets
+      .empty()
+      .en('Very happy milk produced by very happy cow!')
+      .de('Sehr glückliche Milch von sehr glücklicher Kuh!');
+
+    const slug = LocalizedString.presets.empty().en('happy-cow-milk-slug');
+
+    return ProductProjectionRest.random()
+      .productType(ProductType.presets.milk())
```
tdeekens, line 33: This is initially where my will to share code triggers. Rightfully or not. This will "blow" presets up into two variations from one, right?

core/src/builder.ts
CarlosCortizasCT: That is the idea, yes.

models/order-alternative/src/generators.ts
CarlosCortizasCT: But that's not required, isn't it?

models/order-alternative/src/index.ts
CarlosCortizasCT: That's an interesting alternative we must evaluate, although I personally find it easier for the package to export its assets form a main entry point.
Open for discussion.

models/order-alternative/src/order-from-cart-draft/generators.ts
CarlosCortizasCT: At this point, I'm not sure if they always are the same. If they are, I agree it would be easier to maintain just one, although perhaps a little more tricky to understand from the consumer side why we only have one draft model representation and two query model ones.

models/product-projection-alternative/src/presets/happy-cow-milk-product-projection.ts
CarlosCortizasCT: Yes. With this approach we will need to duplicate the presets for the models where the presets themselves populate model properties which have different type across the two representations.

models/order-alternative/src/generators.ts
tdeekens: Nope. Unless you request the field and it should return an "empty value".

models/order-alternative/src/index.ts
tdeekens: Yes, it's just minor in the end and depends on the consumer I think.

models/order-alternative/src/order-from-cart-draft/generators.ts
tdeekens: > consumer side why we only have one draft model representation and two query model ones.

That's fair but easy to document and should outweigh the "cost" of exposing two - subjectively.

models/product-projection-alternative/src/presets/happy-cow-milk-product-projection.ts
tdeekens: Could be maybe solved though if we could e.g. hydrate or seed a builder with a format which works for either or. Throwing 🍝 at the wall but let's assume we can have where we say `[ ['productType', ProductType.presets.milk()], ['slug', 'happy-cow-milk-slug'] ]` in tuples like `Object.fromObject` we could share the commonality between builder types and reuse it. 

Sort of wondering if that would remove the whole proxy in the end. A bit too wide of a scope I suppose.

CarlosCortizasCT:
This PR was the foundation for other smaller ones where we applied the changes here proposed.

PR #646, opened 2024-08-23
FEC-66: fixes for product type and product data models

dogayuksel:
Fixes three minor issues, can be reviewed commit by commit.

Would enable migration on this PR: https://github.com/commercetools/merchant-center-frontend/pull/17463

PR #647, opened 2024-08-23
FEC-66: Add missing attribute definition type models

dogayuksel:
merge after: https://github.com/commercetools/test-data/pull/646

Would enable migration on this PR: https://github.com/commercetools/merchant-center-frontend/pull/17463

### DIFF
--- a/models/product-type/src/attribute-money-type/attribute-money-type-draft/transformers.ts
+++ b/models/product-type/src/attribute-money-type/attribute-money-type-draft/transformers.ts
@@ -1,0 +1,34 @@
+import { Transformer } from '@commercetools-test-data/core';
+import { TAttributeEnumType } from '../../attribute-enum-type';
+import {
+  TAttributeMoneyTypeDraftGraphql,
+  TAttributeMoneyTypeDraft,
+} from '../types';
+
+const transformers = {
+  default: Transformer<TAttributeMoneyTypeDraft, TAttributeMoneyTypeDraft>(
+    'default',
+    {
+      buildFields: [],
+    }
+  ),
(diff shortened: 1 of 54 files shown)


models/product-type/src/attribute-money-type/attribute-money-type-draft/transformers.ts
```diff
@@ -0,0 +1,34 @@
...
+    {
+      buildFields: [],
+    }
+  ),
+  graphql: Transformer<
+    TAttributeMoneyTypeDraft,
+    TAttributeMoneyTypeDraftGraphql
+  >('graphql', {
+    buildFields: [],
+    replaceFields: () => ({
```
CarlosCortizasCT, line 26: Maybe I'm missing something but here in the transformer we are replacing the `money` property of the draft model but I don't see that property in the generator so I don't understand where that property comes from and also why we are setting it to null here.

Could you maybe explain a bit this part? 🙏

dogayuksel: This is how it is modelled on GraphQL side. Since can't send an empty prop, has a dummy object inside. 

Screenshot 2024-08-26 at 10 23 34

Screenshot 2024-08-26 at 10 23 38

CarlosCortizasCT: Ok. After some digging in the docs and the GraphQL explorer I think I get it.

CarlosCortizasCT (approved):
Thanks! 👍

antoniolodias (approved):
💯

jaikumar-tj (approved):

PR #652, opened 2024-08-27
FEC-39: Create `Attribute-group` model

jaikumar-tj:
### Summary

A new model `attribute-group` is created based on the existing test-data in the FE repository. 

Note: The parent PR is in draft mode due to the `inventory-entry` update being done in a separate PR. This PR can be reviewed independently. I will merge this PR only after parent PR is unblocked and reviewed.

### DIFF
--- a/models/attribute-group/README.md
+++ b/models/attribute-group/README.md
@@ -11,3 +11,12 @@ $ pnpm add -D @commercetools-test-data/attribute-group
 ```
 
 # Usage
+
+```ts
+import {
+  AttributeGroup,
+  type TAttributeGroup,
+} from '@commercetools-test-data/attribute-group';
+
+const attributeGroup = AttributeGroup.random().build<TAttributeGroup>();
+```
--- a/models/attribute-group/README.md
+++ b/models/attribute-group/README.md
@@ -16,7 +16,13 @@ $ pnpm add -D @commercetools-test-data/attribute-group
 import {
   AttributeGroup,
   type TAttributeGroup,
+  type TAttributeGroupGraphql,
 } from '@commercetools-test-data/attribute-group';
 
 const attributeGroup = AttributeGroup.random().build<TAttributeGroup>();
+// For REST entities
+const attributeGroupRest = AttributeGroup.random().buildRest<TAttributeGroup>();
+// For Graphql entities
+const attributeGroupGraphql =
+  AttributeGroup.random().buildGraphql<TAttributeGroupGraphql>();
 ```
--- a/models/attribute-group/README.md
+++ b/models/attribute-group/README.md
@@ -1,0 +1,13 @@
+# @commercetools-test-data/attribute-group
+
+This package provides the data model for the commercetools platform `attributeGroup` representations
+
+https://docs.commercetools.com/api/projects/attribute-groups#attributegroup
+
+# Install
+
+```bash
+$ pnpm add -D @commercetools-test-data/attribute-group
+```
+
+# Usage
--- a/models/attribute-group/src/builder.spec.ts
+++ b/models/attribute-group/src/builder.spec.ts
@@ -14,7 +14,13 @@ describe('builder', () => {
         id: expect.any(String),
         version: expect.any(Number),
         createdAt: expect.any(String),
+        createdBy: expect.objectContaining({
+          customerRef: expect.objectContaining({ typeId: 'customer' }),
+        }),
         lastModifiedAt: expect.any(String),
+        lastModifiedBy: expect.objectContaining({
+          customerRef: expect.objectContaining({ typeId: 'customer' }),
+        }),
         name: expect.any(String),
         nameAllLocales: expect.arrayContaining([
           expect.objectContaining({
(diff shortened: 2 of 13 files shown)


models/attribute-group/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+# Usage
+
+```ts
+import {
+  AttributeGroup,
+  type TAttributeGroup,
+} from '@commercetools-test-data/attribute-group';
+
+const attributeGroup = AttributeGroup.random().build<TAttributeGroup>();
```
CarlosCortizasCT, line 21: Could we please add the rest and graphql examples as well?

models/attribute-group/src/transformers.ts
```diff
@@ -0,0 +1,24 @@
+import { Transformer } from '@commercetools-test-data/core';
+import type { TAttributeGroup, TAttributeGroupGraphql } from './types';
+const transformers = {
+  graphql: Transformer<TAttributeGroup, TAttributeGroupGraphql>('graphql', {
```
CarlosCortizasCT, line 4: Don't we need the rest transformer so the `createdBy`, `lastModifiedBy`, `name` and `description` properties are actually built for that representation?

models/attribute-group/src/generator.ts
```diff
@@ -0,0 +1,23 @@
...
+const [getOlderDate, getNewerDate] = createRelatedDates();
+
+const generator = Generator<TAttributeGroup>({
+  fields: {
+    id: fake((f) => f.string.uuid()),
+    version: sequence(),
+    createdAt: fake(getOlderDate),
+    lastModifiedAt: fake(getNewerDate),
+    name: fake((f) => f.lorem.words(2)),
+    description: fake((f) => f.lorem.sentences(2)),
```
CarlosCortizasCT, line 15: These properties are actually `LocalizedString` se we should be returning one of those models.

models/attribute-group/src/transformers.ts
```diff
@@ -0,0 +1,24 @@
...
+    buildFields: ['createdBy', 'lastModifiedBy'],
+    addFields: ({ fields }) => ({
+      __typename: 'AttributeGroup',
+      nameAllLocales: [
+        {
+          locale: 'en',
+          value: fields.name,
```
CarlosCortizasCT, line 11: `name` and `description` should be `LocalizedString` types so we would need to use that model helpers (`toLocalizedField`) in order to transform them into `LocalizedFields`.
We also need those properties to be buildable.

models/attribute-group/src/builder.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+/* eslint-disable jest/valid-title */
+import { createBuilderSpec } from '@commercetools-test-data/core/test-utils';
+import type { TAttributeGroup, TAttributeGroupGraphql } from './types';
+import * as AttributeGroup from './index';
+
+describe('builder', () => {
+  it(
+    ...createBuilderSpec<TAttributeGroup, TAttributeGroupGraphql>(
+      'graphql',
+      AttributeGroup.random(),
```
CarlosCortizasCT, line 11: The generator populates all the properties but the `attributes` one.
It would be great if we can add them in the test.

models/attribute-group/src/transformers.ts
jaikumar-tj: When we (Pacman) worked on the attribute group, we decided to consume only the GraphQL in FE. I didn't see any benefit in creating rest since it is not used. Do you think for the consistency, every model should have both ?

jaikumar-tj: Done in https://github.com/commercetools/test-data/pull/652/commits/ac5cfe5c389dddd194f3880105d4b470b2da6899

models/attribute-group/src/generator.ts
jaikumar-tj: Done in https://github.com/commercetools/test-data/pull/652/commits/ac5cfe5c389dddd194f3880105d4b470b2da6899

models/attribute-group/src/transformers.ts
jaikumar-tj: Done in https://github.com/commercetools/test-data/pull/652/commits/ac5cfe5c389dddd194f3880105d4b470b2da6899

models/attribute-group/src/builder.spec.ts
jaikumar-tj: Done in https://github.com/commercetools/test-data/pull/652/commits/ac5cfe5c389dddd194f3880105d4b470b2da6899

models/attribute-group/src/transformers.ts
CarlosCortizasCT: Thanks for addressing this comment.

Yes, I think we should keep the consistency in all models. Also because this is a public repo and other consumers might need to use the rest representation.

CarlosCortizasCT (approved):
Looks good to me 👍 

Thanks! 🙇

antoniolodias (approved):

mario-priceless (approved):

PR #654, opened 2024-08-27
FEC-56: Add `CustomersSearchListMyView` my view model

jaikumar-tj:
### Summary
A new model `CustomersSearchListMyView` is created based on the existing test-data in the FE repository. This model is related to internal [MC-settings graphql](https://github.com/commercetools/merchant-center-services/blob/main/packages/settings/lib/schemas/customers-search-list-my-view.graphql)

### DIFF
--- a/models/customers-search-list-my-view/README.md
+++ b/models/customers-search-list-my-view/README.md
@@ -15,9 +15,9 @@ $ pnpm add -D @commercetools-test-data/customers-search-list-my-view
 ```ts
 import type {
   CustomersSearchListMyView,
-  type TCustomersSearchListMyView,
+  type TCustomersSearchListMyViewGraphql,
 } from '@commercetools-test-data/customers-search-list-my-view';
 
 const customersSearchListMyView =
-  CustomersSearchListMyView.random().build<TCustomersSearchListMyView>();
+  CustomersSearchListMyView.random().buildGraphl<TCustomersSearchListMyViewGraphql>();
 ```
--- a/models/customers-search-list-my-view/README.md
+++ b/models/customers-search-list-my-view/README.md
@@ -1,0 +1,23 @@
+# @commercetools-test-data/customers-search-list-my-view
+
+This package provides the data model for the commercetools internal API `CustomersSearchListMyView` type
+
+https://github.com/commercetools/merchant-center-services/blob/main/packages/settings/lib/schemas/customers-search-list-my-view.graphql
+
+# Install
+
+```bash
+$ pnpm add -D @commercetools-test-data/customers-search-list-my-view
+```
+
+# Usage
+
+```ts
+import type {
+  CustomersSearchListMyView,
+  type TCustomersSearchListMyView,
+} from '@commercetools-test-data/customers-search-list-my-view';
+
+const customersSearchListMyView =
+  CustomersSearchListMyView.random().build<TCustomersSearchListMyView>();
+```
--- a/models/customers-search-list-my-view/src/builder.spec.ts
+++ b/models/customers-search-list-my-view/src/builder.spec.ts
@@ -45,6 +45,7 @@ describe('building as GraphQL', () => {
   it('should remove `name`', () => {
     const built: TCustomersSearchListMyViewGraphql =
       CustomersSearchListMyView().buildGraphql();
+    // @ts-ignore
     expect(built.name).not.toBeDefined();
   });
 });
--- a/models/customers-search-list-my-view/src/builder.spec.ts
+++ b/models/customers-search-list-my-view/src/builder.spec.ts
@@ -1,14 +1,11 @@
-import {
-  LocalizedString,
-  LocalizedField,
-} from '@commercetools-test-data/commons';
+import { LocalizedField } from '@commercetools-test-data/commons';
 import { FilterValues } from '@commercetools-test-data/filter-values';
 import {
   TMcSettingsFilterValues,
   TMcSettingsSortOrder,
 } from '@commercetools-test-data/graphql-types';
 import CustomersSearchListMyView from './builder';
-import { TCustomersSearchListMyViewGraphql } from './types';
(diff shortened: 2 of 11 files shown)


models/customers-search-list-my-view/src/types.ts
```diff
@@ -0,0 +1,21 @@
...
+export type TCustomersSearchListMyView =
+  TMcSettingsCustomersSearchListMyView & {
+    name: string;
+  };
+
+export type TCustomersSearchListMyViewGraphql = Omit<
+  TCustomersSearchListMyView,
+  'nameAllLocales'
+> & {
+  nameAllLocales: TLocalizedStringGraphql | null;
```
CarlosCortizasCT, line 14: Do we actually need to overwrite this property?
I'm asking because the type in the `TCustomersSearchListMyView` is already a `LocalizedField`, which is what the Graphql API will return.

Also, in the original type it's defined as an array.

models/customers-search-list-my-view/README.md
```diff
@@ -0,0 +1,23 @@
...
+# Usage
+
+```ts
+import type {
+  CustomersSearchListMyView,
+  type TCustomersSearchListMyView,
+} from '@commercetools-test-data/customers-search-list-my-view';
+
+const customersSearchListMyView =
+  CustomersSearchListMyView.random().build<TCustomersSearchListMyView>();
```
CarlosCortizasCT, line 22: Since we only have a GraphQL representation of this model, perhaps we can use the `buildGraphql` function in the example so readers will know how the expected call will look like:
```suggestion
  CustomersSearchListMyView.random().buildGraphl<TCustomersSearchListMyViewGraphql>();
```

models/customers-search-list-my-view/src/builder.spec.ts
```diff
@@ -0,0 +1,41 @@
+import CustomersSearchListMyView from './builder';
+import { TCustomersSearchListMyViewGraphql } from './types';
+
+describe('building', () => {
```
CarlosCortizasCT, line 4: It would be great if we can write more assertions in the tests.
We already found some inconsistencies in some of the models because their tests were too shallow so I think it's helpful if we try to populate all the model's properties and assert them.

[Here](https://github.com/commercetools/test-data/blob/main/models/product-projection/src/builder.spec.ts#L249) is an example of what I mean.

jaikumar-tj: Done https://github.com/commercetools/test-data/pull/654/commits/29a5a7e24f2bea40ab932efc3ada8db4b9edb70f. Let me know if this is enough

models/customers-search-list-my-view/README.md
jaikumar-tj: Done https://github.com/commercetools/test-data/pull/654/commits/29a5a7e24f2bea40ab932efc3ada8db4b9edb70f

models/customers-search-list-my-view/src/types.ts
jaikumar-tj: Done https://github.com/commercetools/test-data/pull/654/commits/29a5a7e24f2bea40ab932efc3ada8db4b9edb70f

models/customers-search-list-my-view/src/builder.spec.ts
CarlosCortizasCT: Yes. Thanks! 👍

models/customers-search-list-my-view/src/types.ts
```diff
@@ -0,0 +1,16 @@
+import { TBuilder } from '@commercetools-test-data/core';
+import type { TMcSettingsCustomersSearchListMyView } from '@commercetools-test-data/graphql-types';
+
+export type TCustomersSearchListMyView =
+  TMcSettingsCustomersSearchListMyView & {
+    name: string;
+  };
+
+export type TCustomersSearchListMyViewGraphql = TCustomersSearchListMyView & {
+  __typename: 'CustomersSearchListMyView';
```
CarlosCortizasCT, line 10: This property is already declared in the type imported from the `@commercetools-test-data/graphql-types` package ([reference](https://github.com/commercetools/test-data/blob/main/graphql-types/src/generated/settings.ts#L485)).

jaikumar-tj: 🔍 You got an eagle's eye.  Fixed in https://github.com/commercetools/test-data/pull/654/commits/384d66ea09f895d0ea3607ffc7f1657b4b9f5781

models/customers-search-list-my-view/src/generator.ts
```diff
@@ -0,0 +1,57 @@
...
+    userId: fake((f) => f.string.uuid()),
+    /**
+     * projectKey: String!
+     */
+    projectKey: fake((f) => f.lorem.slug(2)),
+    /**
+     * nameAllLocales [LocalizedField!]
+     */
+    nameAllLocales: null,
+    name: fake(() => LocalizedString.random()),
```
CarlosCortizasCT, line 35: I'm sorry but I don't fully understand this property because this model only exists in the GraphQL representation and it does not have this property.
Could you maybe explain why we need it? 🙏 

image

jaikumar-tj: Good point! I'm not sure why this field was implemented in this [model here](https://github.com/commercetools/merchant-center-frontend/blob/main/packages-shared/test-data/src/customers-search-list-my-view/generator.js). I'll take a look at the usage and see if i find something. Otherwise I'll update the generator as per graphql

jaikumar-tj: Updated https://github.com/commercetools/test-data/pull/654/commits/621ac54137f978068e064114a6fb6754b3482b38

CarlosCortizasCT (approved):
Looks good to me 👍 

Thanks! 🙇

PR #655, opened 2024-08-28
FEC-57: Add `platform-limits` model (Part 1)

jaikumar-tj:
### Summary

This is the 1st PR in this series to add the `platform-limits` model. I couldn't find any sdk related to platform-limits for defining typescript types, instead I hardcoded the type based on individual model.

The models are copied from FE repo and few adjustment to `spec` file and `typescript` are done. The model covered in this PR are referred from this below code
- [business-unit](https://github.com/commercetools/merchant-center-frontend/tree/main/packages-shared/test-data/src/platform-limits/business-units)
- [limit](https://github.com/commercetools/merchant-center-frontend/tree/main/packages-shared/test-data/src/platform-limits/limit)
- [limit-with-current](https://github.com/commercetools/merchant-center-frontend/tree/main/packages-shared/test-data/src/platform-limits/limit-with-current)


models/platform-limits/src/business-units/types.ts
```diff
@@ -0,0 +1,19 @@
+import type { TBuilder } from '@commercetools-test-data/core';
+import type { TLimit } from '../limit';
+
+export type TBusinessUnitLimitsProjection = {
+  maxDivisions: TLimit;
+  maxDepthLimit: TLimit;
+  maxAssociates: TLimit;
+  maxAssociateRoles: TLimit;
+};
```
jaikumar-tj, line 9: I couldn't find the `type` in any of the repo. I wrote them down based on the schema from graphql explorer
Screenshot 2024-08-28 at 08 44 19

models/platform-limits/src/index.ts
```diff
@@ -0,0 +1,7 @@
+// Export types
+export * from './business-units/types';
+export * from './limit-with-current/types';
+
+// Export models
+export * as BusinessUnitLimitsProjection from './business-units';
```
jaikumar-tj, line 6: should i change the folder name also reflect the API model?
`business-unit-limits-projection` ? or is it OK to keep it same as in FE test-data?

models/platform-limits/README.md
```diff
@@ -0,0 +1,17 @@
...
+
+```bash
+$ pnpm add -D @commercetools-test-data/platform-limits
+```
+
+# Usage
+
+```ts
+
+```
```
jaikumar-tj, line 17: Readme is updated in the [part-3 PR](https://github.com/commercetools/test-data/pull/657)

models/platform-limits/src/business-units/builder.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+import * as BusinessUnitsPlatformLimits from './index';
+
+describe('building', () => {
+  it(
+    ...createBuilderSpec<
+      TBusinessUnitLimitsProjection,
+      TBusinessUnitLimitsProjection
+    >(
+      'default',
+      BusinessUnitsPlatformLimits.random(),
```
CarlosCortizasCT, line 14: Here we're testing the default values of the generator, which does not seem to provide too much value since all properties are initialized with `null`.
Perhaps we can use the `withLimit` preset so we make sure the model is instantiated with some actual data.

models/platform-limits/src/business-units/transformers.ts
```diff
@@ -0,0 +1,46 @@
...
+const transformers = {
+  default: Transformer<
+    TBusinessUnitLimitsProjection,
+    TBusinessUnitLimitsProjection
+  >('default', {
+    buildFields: [
+      'maxDivisions',
+      'maxDepthLimit',
+      'maxAssociates',
+      'maxAssociateRoles',
```
CarlosCortizasCT, line 16: **nit**: we could extract this list to a constant in the file and reuse it in the three transformers.

models/platform-limits/src/index.ts
CarlosCortizasCT: I vote for keeping it like it currently is as the nested folder nature already tell us this is a platform limit

models/platform-limits/src/limit-with-current/presets/with-warning-exceeded.ts
```diff
@@ -0,0 +1,5 @@
+import LimitWithCurrent from '../builder';
+
+const withWarningExceeded = () => LimitWithCurrent().limit(10).current(8);
```
CarlosCortizasCT, line 3: Shouldn't `current` be a greater number than `limit`?

models/platform-limits/src/limit-with-current/presets/with-exceeded.ts
```diff
@@ -0,0 +1,5 @@
+import LimitWithCurrent from '../builder';
+
+const withExceeded = () => LimitWithCurrent().limit(10).current(10);
```
CarlosCortizasCT, line 3: Shouldn't `current` be a greater number than `limit`?

models/platform-limits/src/limit-with-current/generator.ts
```diff
@@ -0,0 +1,15 @@
...
+import { TLimitWithCurrent } from './types';
+
+/**
+ * limit - Number
+ * current - Number
+ */
+const generator = Generator<TLimitWithCurrent>({
+  fields: {
+    limit: fake((f) => f.number.int(10)),
+    current: fake((f) => f.number.int(10)),
```
CarlosCortizasCT, line 11: Shouldn't `current` be a lower number than `limit` in the generator? (default value)

models/platform-limits/src/business-units/types.ts
CarlosCortizasCT: Fair enough 👍

models/platform-limits/src/limit-with-current/presets/with-exceeded.ts
jaikumar-tj: Based on this [util file](https://github.com/commercetools/merchant-center-frontend/blob/main/packages-shared/experimental-components/src/features/platform-limits/shared/utils/utils.js), if the both value are same, then it already exceeded

models/platform-limits/src/limit-with-current/generator.ts
jaikumar-tj: They both can have same number. Refer above comment.

models/platform-limits/src/business-units/transformers.ts
jaikumar-tj: done https://github.com/commercetools/test-data/pull/655/commits/d01351de3bd1ac841ec8139721c78efb1fbcad99

models/platform-limits/src/business-units/builder.spec.ts
jaikumar-tj: done https://github.com/commercetools/test-data/pull/655/commits/617107cad9f4c095a0f686bf60262289367ca677

models/platform-limits/src/limit-with-current/generator.ts
CarlosCortizasCT: Thanks for the explanation.

What I mean is that, if they both have the same number, the limit will be considered exceeded, right?
If that's correct, then the default value for this limit that the model will be returning will be in the exceeded state, where perhaps it makes more sense that the default value is an allowed one.

Would that make sense to you?

models/platform-limits/src/limit-with-current/presets/with-warning-exceeded.ts
CarlosCortizasCT: Thanks for the explanation.
I understand the rule you mention, but readers coming to this file will not know about it.

For maintainability purposes, I think it would be easier to understand if we use a larger value. As an alternative, we can also leave the `8` but add a comment about the 80% rule.

models/platform-limits/src/limit-with-current/generator.ts
jaikumar-tj: I was thinking that by keeping the default state with same maximum value can generate all the different possible states (eg: not exceeded, with warning exceeded, exceeded). Since we already have a preset for all the different states, if user prefer only `allowed` one or `exceeded one` then they can choose the corresponding preset. 

Do you think we make default to generate only `not-exceeded` value? or both `not-exceeded` & `with-warning-exceeded` (because both can be considered as `allowed` one?
In other words, do you prefer the current is `int(9)` or `int(7)` or something else?

CarlosCortizasCT: My vote wold be for `int(7)` as the default one and users customizing that value or using a preset if they want something else.

jaikumar-tj: Done https://github.com/commercetools/test-data/pull/655/commits/aa7ad4e28b70b1fa5e0b1d52d4c63aaa99db10d1

antoniolodias (approved):

models/platform-limits/src/limit-with-current/presets/with-warning-exceeded.ts
jaikumar-tj: Added comment https://github.com/commercetools/test-data/pull/655/commits/fcd86866594399f9f3a762d5eaa7b027d81aa686

CarlosCortizasCT (approved):
Thanks! 👍

PR #666, opened 2024-09-03
[FEC-49] - add enum value, localized enum value and field type models

antoniolodias:
#### Summary

This PR creates `enum value`, `localized enum value` and `field type` models needed for this [migration PR](https://github.com/commercetools/merchant-center-frontend/pull/17434) which is based on [batch 12](https://commercetools.atlassian.net/browse/FEC-49) of the [chapter epic](https://commercetools.atlassian.net/browse/FEC-26) where we are migrating tests to use public test-data package


models/type/src/field-type/transformers.ts
```diff
@@ -0,0 +1,90 @@
...
+  } as T;
+
+  switch (fields.name) {
+    case 'Enum':
+      return {
+        ...commonFields,
+        values: buildFields(
+          (fields.values as TCustomFieldEnumValue[]) ||
+            list(CustomFieldEnumValue.random),
+          'rest'
```
jaikumar-tj, line 36: shouldn't this type `rest` be based on the caller? I mean, this `resolveCommonTransformations` is called by all 3 transformers below

models/type/src/field-type/transformers.ts
```diff
@@ -0,0 +1,90 @@
...
+import * as CustomFieldLocalizedEnumValue from '../custom-field-localized-enum-value';
+import { TCustomFieldLocalizedEnumValue } from '../custom-field-localized-enum-value';
+import FieldType from './builder';
+import { typeNames, graphqlTypenameByFieldTypeName } from './constants';
+import { TFieldName, TFieldType, TFieldTypeGraphql } from './types';
+
+const list = <T>(fn: (index?: number) => T) =>
+  Array.from({ length: faker.number.int({ min: 1, max: 7 }) }, (_, index) =>
+    fn(index)
+  );
```
jaikumar-tj, line 19: Could you maybe add a comment explaining what this function is doing? I see `min` and `max` and index being passed to the callback function, but not fully understand its function.

models/type/src/localized-enum-value/types.ts
```diff
@@ -0,0 +1,15 @@
+import { TLocalizedFieldGraphql } from '@commercetools-test-data/commons';
+import type { TBuilder } from '@commercetools-test-data/core';
+
+export type TLocalizedEnumValue = {
+  key: string;
+  label: string;
```
CarlosCortizasCT, line 6: Could it be that this is of type `LocalizedString` in the REST representation?

models/type/src/field-type/transformers.ts
```diff
@@ -0,0 +1,90 @@
...
+
+const resolveCommonTransformations = <T extends TFieldType | TFieldTypeGraphql>(
+  fields: TFieldType
+): T => {
+  const commonFields = {
+    name: fields.name as TFieldName,
+    __typename: graphqlTypenameByFieldTypeName[fields.name],
```
CarlosCortizasCT, line 26: This property should not be common to all transformers, right? I think it's supposed to only exists in the GraphQL representation.

models/type/src/field-type/transformers.ts
CarlosCortizasCT: Updated here: 561d23675c46fbf0307cc2f8c0fa2696220ede83

models/type/src/field-type/transformers.ts
CarlosCortizasCT: Actually, the index was not needed.

This is just used to create an array of a model builder between 1 and 7 elements (what was used before the migration).

Added a comment here: 561d23675c46fbf0307cc2f8c0fa2696220ede83

CarlosCortizasCT (approved):
Awesome 🚀 

Thanks Antonio 👍

jaikumar-tj (approved):

PR #669, opened 2024-09-03
[FEC-49] - improve order model and build needed sub models

antoniolodias:
#### Summary

This PR improves the `order` model and creates `delivery`, `delivery item`, `line item return item`, `parcel`, `parcel measurements`, `return info` and `tracking data`  models needed for this [migration PR](https://github.com/commercetools/merchant-center-frontend/pull/17434) which is based on [batch 12](https://commercetools.atlassian.net/browse/FEC-49) of the [chapter epic](https://commercetools.atlassian.net/browse/FEC-26) where we are migrating tests to use public test-data package


models/order/src/delivery/transformers.ts
```diff
@@ -0,0 +1,19 @@
+import { Transformer } from '@commercetools-test-data/core';
+import type { TDelivery, TDeliveryGraphql } from './types';
+
+const transformers = {
+  default: Transformer<TDelivery, TDelivery>('default', {
+    buildFields: ['items', 'parcels', 'address'],
```
jaikumar-tj, line 6: I think `custom` should be added here, incase if the customfield is implemented then that model has to be built. Similar to how it is written in below  `models/order/src/parcel/transformers.ts`

Same for other transformers below like `models/order/src/line-item-return-item/transformers.ts`, etc

models/order/src/return-info/builder.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+
+// TODO: fix whatever is happening to the items field
+describe('builder', () => {
+  const defaultReturnInfoSpec = {
+    // items: expect.arrayContaining([
+    //   expect.objectContaining({
+    //     id: expect.any(String),
+    //     quantity: expect.any(Number),
+    //   }),
+    // ]),
```
jaikumar-tj, line 16: what was the issue happening with `items`? is it just typescript issue or something else?

models/order/src/return-info/transformers.ts
```diff
@@ -0,0 +1,19 @@
...
+
+const transformers = {
+  default: Transformer<TReturnInfo, TReturnInfo>('default', {
+    buildFields: [],
+  }),
+  rest: Transformer<TReturnInfo, TReturnInfo>('rest', {
+    buildFields: [],
+  }),
+  graphql: Transformer<TReturnInfo, TReturnInfoGraphql>('graphql', {
+    buildFields: [],
```
jaikumar-tj, line 12: I think `items` needs to be built. I assume this was the reason for the issue in the builder.spec file above where you left a comment

models/order/src/builder.spec.ts
```diff
@@ -303,20 +186,22 @@ describe('builder', () => {
...
+          customerRef: expect.objectContaining({ typeId: 'customer' }),
+          userRef: expect.objectContaining({ typeId: 'user' }),
+        }),
         __typename: 'Order',
       })
     )
   );
 });
+
+//TODO: add tests when customizing
```
CarlosCortizasCT, line 207: Instead of adding more tests, perhaps it's easier if we just extend the ones we already have to provide not only the Order with the default values but also configure the optional fields.

models/order/src/delivery-item/transformers.ts
```diff
@@ -0,0 +1,21 @@
...
+const transformers = {
+  default: Transformer<TDeliveryItem, TDeliveryItem>('default', {
+    buildFields: [],
+  }),
+  rest: Transformer<TDeliveryItem, TDeliveryItem>('rest', {
+    buildFields: [],
+  }),
+  graphql: Transformer<TDeliveryItem, TDeliveryItemGraphql>('graphql', {
+    buildFields: [],
+    replaceFields: ({ fields }) => ({
```
CarlosCortizasCT, line 13: Since we're not replacing any model's field here, it would be better to use the `addFields` property instead. Doing so will allow us to not need to return the whole fields but just the ones we're adding.

models/order/src/delivery/builder.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+    const parcel = Parcel.random().id(id);
+    const address = Address.random().streetName(streetName);
+
+    const deliveryMock = Delivery.random()
+      .id(id)
+      .key(key)
+      .items([item])
+      .parcels([parcel])
+      .address(address)
+      // .custom()
```
CarlosCortizasCT, line 67: It seems we forgot to populate the `custom` property.

models/order/src/line-item-return-item/builder.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+
+    const lineItemReturnItemMock = LineItemReturnItem.random()
+      .id(id)
+      .key(key)
+      .quantity(quantity)
+      .comment(comment)
+      .shipmentState(shipmentStateAdvised)
+      .paymentState(paymentStateNonRefundable)
+      .lineItemId(lineItemId)
+      // .custom(custom)
```
CarlosCortizasCT, line 65: We're missing this property.

models/order/src/line-item-return-item/transformers.ts
```diff
@@ -0,0 +1,22 @@
+import { Transformer } from '@commercetools-test-data/core';
+import type { TLineItemReturnItem, TLineItemReturnItemGraphql } from './types';
+
+const transformers = {
```
CarlosCortizasCT, line 4: It seems we're missing the `custom` property in the `buildFields` list of the transformers.

models/order/src/parcel/builder.spec.ts
```diff
@@ -0,0 +1,95 @@
...
+    const trackingData = TrackingData.random().carrier(carrier);
+    const deliveryItem = DeliveryItem.random().quantity(quantity);
+
+    const parcelMock = Parcel.random()
+      .id(id)
+      .key(key)
+      .measurements(measurements)
+      .trackingData(trackingData)
+      .items([deliveryItem])
+      // .custom(custom)
```
CarlosCortizasCT, line 73: We're missing the configuration for this property.

models/order/src/parcel/transformers.ts
```diff
@@ -0,0 +1,19 @@
+import { Transformer } from '@commercetools-test-data/core';
+import type { TParcel, TParcelGraphql } from './types';
+
+const transformers = {
+  default: Transformer<TParcel, TParcel>('default', {
+    buildFields: ['items', 'custom', 'measurements', 'trackingData'],
```
CarlosCortizasCT, line 6: Since we're using the same list in all transformers, perhaps we can extract it to a constant.

models/order/src/return-info/transformers.ts
antoniolodias: [228273d](https://github.com/commercetools/test-data/pull/669/commits/228273d07306b97f0119f2a1645f3557186851ad) thank you! I also forgot to fake it inside the generator :see_no_evil:

models/order/src/delivery-item/transformers.ts
```diff
@@ -0,0 +1,21 @@
...
+      ...fields,
+      lineItemId: fields.id,
```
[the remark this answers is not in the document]
antoniolodias: my mistake.... thank you

models/order/src/parcel/builder.spec.ts
CarlosCortizasCT: After discussing about this property with @antoniolodias, we agree that some more missing models need to be migrated, making the scope of this PR even bigger.

Since this property is not currently used in the MC local test data model we're bringing in this PR, we agreed to keep that property commented by now, in order to unblock this PR and others that depend on this one, and work on it in a [separate task](https://commercetools.atlassian.net/browse/FEC-71).

models/order/src/builder.spec.ts
CarlosCortizasCT: Tests have been updated here: 145fd9bd1b120fefc79d575e80184bae01eb9209

models/order/src/delivery/transformers.ts
CarlosCortizasCT: @jaikumar-tj Please see [this comment](https://github.com/commercetools/test-data/pull/669#discussion_r1745521489) below.

CarlosCortizasCT (approved):
Many thanks, Antonio 🙇

jaikumar-tj (approved):

PR #671, opened 2024-09-04
FEC-60: Fix `state` model and add `CategorySearch` model

jaikumar-tj:
### Summary

The `name` field in state model is transformed for graphql. A new `CategorySearch` model is added under category.

Note: `CategorySearch` is a deprecated graphql service and it is used only in category search page which was developed long time ago.

### DIFF
--- a/models/category/src/category-search/builder.spec.ts
+++ b/models/category/src/category-search/builder.spec.ts
@@ -1,6 +1,7 @@
 /* eslint-disable jest/no-disabled-tests */
 /* eslint-disable jest/valid-title */
 
+import { LocalizedString } from '@commercetools-test-data/commons';
 import { createBuilderSpec } from '@commercetools-test-data/core/test-utils';
 import type { TCategorySearch, TCategorySearchGraphql } from '../types';
 import * as CategorySearch from './index';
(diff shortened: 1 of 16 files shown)


models/category/src/category-search/builder.spec.ts
```diff
@@ -0,0 +1,178 @@
...
+        }),
+        childCount: expect.any(Number),
+        stagedProductCount: expect.any(Number),
+        ancestorsRef: [],
+        ancestors: [],
+        productTypeNames: [],
+        children: [],
+        assets: [],
+        custom: [],
+      })
```
antoniolodias, line 51: nit: I don't see this being done a lot, so I am not sure if there are reasons against it, but personally I think its easier if we set this big block into a variable and reuse it in every test, it helps understand what are the differences between rest and graphQL

antoniolodias (approved):
Looks good!!

models/category/src/category-search/builder.spec.ts
jaikumar-tj: Done https://github.com/commercetools/test-data/pull/671/commits/18bd5cf9b5f4711dd329e87a5bdc7128affff1ec

jocelyn-dunkley (approved):

PR #678, opened 2024-09-06
chore: export field type from type model

antoniolodias:
export field type from type model


jaikumar-tj (approved):
Don't forget to add changeset before merging!

PR #680, opened 2024-09-09
Update Standalone price model

nima-ct:
This PR updates the Standalone price model to improve consistency in our unit tests and provide a more realistic model that aligns with our GraphQL schema and form limitations.
Currencies are synced now with the main value currency code.

### DIFF
--- a/models/standalone-price/src/transformers.ts
+++ b/models/standalone-price/src/transformers.ts
@@ -11,6 +11,19 @@ import type {
 } from './types';
 
 const transformers = {
+  default: Transformer<TStandalonePrice, TStandalonePrice>('default', {
+    buildFields: [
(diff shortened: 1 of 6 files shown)


mario-priceless (approved):

models/standalone-price/src/transformers.ts
```diff
@@ -19,6 +19,9 @@ const transformers = {
       'customerGroup',
       'channel',
       'tiers',
+      'custom',
+      'discounted',
+      'staged',
```
ragafus, line 24: I have no idea what's the difference between `default` and `rest`. Should we also add `replaceFields` here?

ragafus (approved):

PR #681, opened 2024-09-10
chore: add jest-extended to allow more specific matchers

ragafus:
This pull request adds `jest-extended` to the project to allow more specific matchers in the tests.

[jest-extended](https://github.com/jest-community/jest-extended) is already [used in the MC monorepo](https://github.com/commercetools/merchant-center-frontend/blob/main/jest.test.config.js#L35)

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -12,6 +12,5 @@
     "isolatedModules": true,
     "resolveJsonModule": true,
     "allowJs": false
-  },
-  "typeRoots": ["./node_modules/@types", "./global.d.ts"]
+  }
 }


tsconfig.json
```diff
@@ -14,6 +11,7 @@
     "forceConsistentCasingInFileNames": true,
     "isolatedModules": true,
     "resolveJsonModule": true,
-    "allowJs": false
+    "allowJs": false,
+    "types": ["jest", "jest-extended"]
```
emmenko, line 15: There should be no need for this, as all other types will be excluded.

Have you checked this maybe? https://jest-extended.jestcommunity.dev/docs/getting-started/typescript

ragafus: > There should be no need for this, as all other types will be excluded.
> 
> Have you checked this maybe? https://jest-extended.jestcommunity.dev/docs/getting-started/typescript

With that approach `preconstruct` fails https://github.com/commercetools/test-data/actions/runs/10812738929/job/29995067358?pr=681#step:7:11

ragafus: > > There should be no need for this, as all other types will be excluded.
> > Have you checked this maybe? https://jest-extended.jestcommunity.dev/docs/getting-started/typescript
> 
> With that approach `preconstruct` fails https://github.com/commercetools/test-data/actions/runs/10812738929/job/29995067358?pr=681#step:7:11

I replaced `"files": ["global.d.ts"]` with `"typeRoots": ["./node_modules/@types", "./global.d.ts"]` and it worked. It doesn't exclude any type present in  `./node_modules/@types` and it adds the needed `jest-extended` types.

emmenko:
@ragafus I pushes a commit to remove the `typeRoots` overwrite. My editor recognizes the matchers via the `global.d.ts` file, there should be no need to explicitly import it.

image

Can you give it a try?

Note: overwriting `types` or `typeRoots` is tricky and most of the time is not necessary, so I would prefer to avoid doing that.

emmenko (approved):

ragafus:
> @ragafus I pushes a commit to remove the `typeRoots` overwrite. My editor recognizes the matchers via the `global.d.ts` file, there should be no need to explicitly import it.
> 
> image
> Can you give it a try?
> 
> Note: overwriting `types` or `typeRoots` is tricky and most of the time is not necessary, so I would prefer to avoid doing that.

You are right. It worked.

PR #683, opened 2024-09-11
feat: add applicationMode field to `CartDiscountValueAbsolute` model

ragafus:
To properly test Cart Discount Application Modes feature, the new field needs to be added to the `CartDiscountValueAbsolute` model.

### DIFF
--- a/models/cart-discount/src/cart-discount-value-absolute/types.ts
+++ b/models/cart-discount/src/cart-discount-value-absolute/types.ts
@@ -7,12 +7,24 @@ import type { TBuilder } from '@commercetools-test-data/core';
 export type TCartDiscountValueAbsolute = CartDiscountValueAbsolute;
 export type TCartDiscountValueAbsoluteDraft = CartDiscountValueAbsoluteDraft;
 
-export type TCartDiscountValueAbsoluteGraphql = TCartDiscountValueAbsolute & {
+export type TCartDiscountValueAbsoluteGraphql = Omit<
+  TCartDiscountValueAbsolute,
+  'applicationMode'
+> & {
   __typename: 'AbsoluteDiscountValue';
 };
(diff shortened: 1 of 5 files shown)


models/cart-discount/src/cart-discount-value-absolute/types.ts
```diff
@@ -7,12 +7,24 @@ import type { TBuilder } from '@commercetools-test-data/core';
...
-export type TCartDiscountValueAbsoluteGraphql = TCartDiscountValueAbsolute & {
+export type TCartDiscountValueAbsoluteGraphql = Omit<
+  TCartDiscountValueAbsolute,
+  'applicationMode'
+> & {
   __typename: 'AbsoluteDiscountValue';
 };
 
+export type TCartDiscountValueAbsoluteCartGraphql =
+  Required<TCartDiscountValueAbsolute> & {
```
nima-ct, line 18: I'm curious about the `Required` utility here. What are the benefits?

nima-ct (approved):

models/cart-discount/src/cart-discount-value-absolute/types.ts
ragafus: > I'm curious about the `Required` utility here. What are the benefits?

`applicationMode` prop is optional in `TCartDiscountValueAbsolute` and mandatory in the GraphQL schema.
IMO we should have a tool in this repo to generate GraphQL types from GraphQL fragments, this way we would not need manual types like this one which could be innacurate.

nima-ct: I agree. Deriving GraphQL types from Rest types doesn't seem ideal

mario-priceless (approved):

PR #687, opened 2024-10-03
New test data models documentation

CarlosCortizasCT:
This PRs add new documentation to the repository where we explain how new test data models look like and how to build them.

PR #688, opened 2024-10-03
Update Core helpers to support the new proposed way of writing test data models

CarlosCortizasCT:
Based on the agreements achieved in [this RFC](https://commercetools.atlassian.net/wiki/spaces/FEC/pages/1068335125/RFC+Test-data+models+refactor+proposal), we're now implementing (step by step) the changes required in the repository in order to support them.

We already have [this PR](https://github.com/commercetools/test-data/pull/687) when we add documentation about this new patterns to follow when writing test data models.

In this PR we're actually updating the core helpers so they can support the current versions of the test data models as well as the new one and also a compatibility mode where some data models need to support both at the same time.


core/src/builder.ts
```diff
@@ -34,8 +30,12 @@ const createState = <Model>({
 
   return {
     get: () => state,
-    merge: (update: Partial<Model>) => {
-      state = { ...state, ...update };
+    merge: (update: Partial<Model>, overwrite = true) => {
```
CarlosCortizasCT, line 33: Changes in this object are required so we can initialize the state with the generator result after the builder is actually built.

We need a state in the builder so chained calls to its properties can be registered but we cannot set the initial state at builder instantiation time as we don't know yet which version of the builder we're going to use at that point (this is because the compatibility mode).

This solution allows to update the state with the generator result later during the actual build call.

core/src/builder.ts
```diff
@@ -77,13 +86,27 @@ function PropertyBuilder<Model>(initialProps?: Partial<Model>) {
 function Builder<Model>({
   generator,
   transformers,
+  type,
```
CarlosCortizasCT, line 89: In the new pattern, we will be receiving which representation this builder is for: rest | graphql

core/src/builder.ts
```diff
@@ -77,13 +86,27 @@ function PropertyBuilder<Model>(initialProps?: Partial<Model>) {
...
+  postBuild,
```
CarlosCortizasCT, line 90: In the new pattern, this is a callback used to manipulate the output of the build before it's returned to the consumer. This allows to resolve dependent properties of the data model.

core/src/builder.ts
```diff
@@ -77,13 +86,27 @@ function PropertyBuilder<Model>(initialProps?: Partial<Model>) {
...
+  name = 'Unknown Builder',
```
CarlosCortizasCT, line 91: This helps debugging a lot.

core/src/builder.ts
```diff
@@ -77,13 +86,27 @@ function PropertyBuilder<Model>(initialProps?: Partial<Model>) {
...
+  compatConfig,
```
CarlosCortizasCT, line 92: This is for the compatibility mode of the new pattern where a single builder needs to support both representations.

core/src/helpers.ts
```diff
@@ -168,6 +172,80 @@ const buildRestList = <Model>(
...
+    }),
+  };
+};
+
+type TCreateSpecializedBuilderParams<TModel> = {
+  modelFieldsConfig: TModelFieldsConfig<TModel>;
+  type: 'rest' | 'graphql';
+  name: string;
+};
+const createSpecializedBuilder = <TModel>(
```
CarlosCortizasCT, line 195: Helper to create the new rest and graphql Builders we will be using with the new pattern.

core/src/helpers.ts
```diff
@@ -168,6 +172,80 @@ const buildRestList = <Model>(
...
+    transformers: createSpecializedTransformers<TModel>({
+      type: params.type,
+    }),
+    postBuild: params.modelFieldsConfig.postBuild,
+  });
+
+  return modelBuilder as TBuilder<TModel>;
+};
+
+const createCompatibilityBuilder = <TModel>(params: {
```
CarlosCortizasCT, line 213: Helper to create the compatibility builder so current data models can be migrated to the new patterns but still (temporary) expose the same external API (backwards compatibility).

core/src/transformer.ts
```diff
@@ -36,6 +43,21 @@ function Transformer<Model, TransformedModel>(
           };
         }
       });
+    } else if (fieldsToBuild !== false) {
```
CarlosCortizasCT, line 46: We now build all buildable data model properties by default so we don't need to specify an array of those ones at the data model level.

core/src/types.ts
```diff
@@ -87,39 +95,58 @@ export type TBuilder<OriginalModel> = {
     OriginalModel[K]
   >;
 } & {
-  build<TransformedModel>(
+  build<TransformedModel = OriginalModel>(
     args?: TFieldBuilderArgs<OriginalModel>
   ): TransformedModel;
-  buildGraphql<TransformedModel>(
+  buildGraphql<TransformedModel = OriginalModel>(
```
CarlosCortizasCT, line 101: This allows for the function to return the expected type by default which works very handy with the new patterns.

models/channel/src/builder.ts
```diff
@@ -6,6 +6,7 @@ import type { TCreateChannelBuilder } from './types';
 
 const Model: TCreateChannelBuilder = () =>
   Builder<Channel>({
+    name: 'Channel',
```
CarlosCortizasCT, line 9: There are a bunch of models where I added the name in order to help debugging.

I think is better to keep them.

models/commons/src/client-logging/transformers.ts
```diff
@@ -12,6 +12,7 @@ const transformers = {
     buildFields: ['customer'],
   }),
   graphql: Transformer<TClientLogging, TClientLoggingGraphql>('graphql', {
+    buildFields: false,
```
CarlosCortizasCT, line 15: Since previously we needed to provide the `buildFields` array in order to specify which data models properties should be automatically built, omitting that property meant build no property at all.

Now that this field is not even needed, for those current cases where we don't actually want to build any property we need to explicitly return `false`.

core/src/builder.spec.ts
```diff
@@ -287,71 +277,6 @@ describe('building', () => {
       );
     });
 
-    it('should build all build upon properties with callbacks', () => {
```
CarlosCortizasCT, line 290: We had a feature that allowed properties updaters to provide a value based on the previous one. 
Keeping this feature with the new proposal is quite hard so, since I haven't found any model in this repo using this feature, I'd like to remove it by now.

If we find any MC applications tests actually using this feature, then I'll get back to it if necessary but I would like to forget about it by now if possible.

core/src/types.ts
```diff
@@ -49,12 +51,17 @@ export type TTransformerOptions<Model, TransformedModel> = {
   addFields?: (args: { fields: Model }) => Partial<TransformedModel>;
   removeFields?: (keyof Model)[];
   replaceFields?: (args: { fields: Model }) => TransformedModel;
-  buildFields?: (keyof Model)[];
+  buildFields?: (keyof Model)[] | false;
```
emmenko, line 54: Nit: let's add a comment to explain what `false` refers to and when it's supposed to be used.

core/src/transformer.ts
emmenko: Nit: Would it make sense to "invert" the logic? If some fields are NOT supposed to be built, we would need to explicitly define them.

If so, we can replace the property `fieldsToBuild` with something like `omitFieldsToBuild`.

emmenko (commented):
Thanks again for tackling this beast! 🙌

I think I was able to follow along with the core changes but I trust you know what you are doing 😉

Overall it looks good to me 🤗

core/src/types.ts
CarlosCortizasCT: Added here: af759dae62574ff496764d3fdcab22bd76f1126b

.changeset/blue-sloths-repair.md
```diff
@@ -0,0 +1,12 @@
...
+'@commercetools-test-data/customer-group': minor
+'@commercetools-test-data/product-type': minor
+'@commercetools-test-data/channel': minor
+'@commercetools-test-data/commons': minor
+'@commercetools-test-data/core': minor
+---
+
+The main change is about the `core` package where we are introducing support for writing test data models using new implementation patterns which makes the process simpler. Also, the resulting code will be more maintainable.
+
+You can head over [here]() for updated documentation about those new patterns.
```
CarlosCortizasCT, line 12: I'll update this link once the [docs PR](https://github.com/commercetools/test-data/pull/687) is merged.

tdeekens (approved):

Sarah4VT (approved):
This is not my area of expertise, but I didn't notice anything I would obviously change.

leventekobor (approved):

PR #689, opened 2024-10-04
New CLI for code generation

CarlosCortizasCT:
This PR introduces a new CLI tool in the repository which helps bootstraping new test data models.

It can create both a new standalone package data model or a child data model to an existing package.

You need to run command (from the repository root's directory):
```bash
pnpm generate-model
```


.husky/commit-msg
```diff
@@ -1,4 +1 @@
-#!/bin/sh
```
CarlosCortizasCT, line 1: These changes in `husky` are not related but I was getting this warning in the console when I was trying to commit new code:
Screenshot 2024-10-04 at 11 12 22

generators/src/new-test-model/templates/package/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+const rest{{modelName}} = {{modelName}}Rest.random().build();
+const graphql{{modelName}} = {{modelName}}Graphql.random().build();
+
+// TODO: Write draft examples if any
+
+// TODO: Write preset examples if any
+// Example
+// const rest{{modelName}}Empty = {{modelName}}Rest.presets.empty().build();
+// const graphql{{modelName}}Empty = {{modelName}}Graphql.presets.empty().build();
```
leventekobor, line 30: @CarlosCortizasCT are you planning to add these TODO's to this PR or a separate one, or these are just leftovers?

leventekobor (approved):

generators/src/new-test-model/templates/package/README.md
CarlosCortizasCT: Thanks for the feedback, Levente 👍 

My idea is that this to keep them. They are part of the TODOs the bootstrapped code will contain to give developers hints about what they should do next.

generators/src/new-test-model/templates/model/types.tpl
```diff
@@ -0,0 +1,27 @@
...
+  {{modelName}},
+} from '@commercetools/platform-sdk';
+import type { TBuilder } from '@commercetools-test-data/core';
+
+export type T{{modelName}}Rest = {{modelName}};
+
+// TODO: Check wether the graphql version has different properties
+// or types than the rest version.
+// You might want to use the GraphQL Explorer for that
+// (https://mc.europe-west1.gcp.escemo.com/test-with-big-data-44/graphql-explorer-commercetools/platform)
```
Sarah4VT, line 11: Is this a url we want in a public repo?

Sarah4VT (approved):
This is great! So useful 🥇 
One question about a url that I just wasn't sure if we should put in the template. Wanted to be sure.

generators/src/new-test-model/templates/model/types.tpl
CarlosCortizasCT: That's a good question.

I think the URL will help developers writing new models but I can see the conflict, specially since it's an _internal_ (Staging) one.

Maybe we can use a production (public) URL or just remove it and hope every engineer knows how to access GraphQL Explorer in the Merchant Center.

@tdeekens What's your take on this one? 🤔

tdeekens: We tried to not expose staging URLs to the public. It's not cause there is something fundamentally risky in doing so. It's just that we try not to so there is no external expectation towards them. In this case we should just use a production URL or a placeholder.

CarlosCortizasCT: Thanks! 👍 

I finally changed it for a placeholder URL:
```
https://<merchant_center_domain>/<project_key>/graphql-explorer-commercetools/platform
```

Updated here: 681e14d45fbee868474df456b29fff910ec97d51

generators/src/new-test-model/new-test-model.generator.ts
```diff
@@ -0,0 +1,102 @@
...
+      modelTemplatesData: modelTemplatesData.map(
+        (template) => template.templatePath
+      ),
+    });
+
+    if (generationType === 'standalone') {
+      packageTemplatesData.forEach((template) => {
+        const filePath = join(outputPath, template.templatePath);
+        ensureDirectory(filePath);
+        writeFileSync(
```
tdeekens, line 76: Should this run through prettier once? Making sure that the generated code is formatted using it according to the rcfile in this repo? We've done this in other CLIs and it's quite useful. In this specific case maybe minor as we rely on the tpl files to be formatted already.

generators/src/new-test-model/templates/model/builders.spec.tpl
```diff
@@ -0,0 +1,27 @@
+import { {{modelName}}Rest, {{modelName}}Graphql } from './index';
```
tdeekens, line 1: Just from a high-level, not that much speaks against using Mustache. Do we need it? We basically do string interpolation without any logic like helpers, partials or logic (like the golden days).

As we use Lodash already and lodash comes with a template function which would maybe be enough already [ref](https://lodash.com/docs/4.17.15#template). Would that also work to avoid using another library?

generators/src/new-test-model/new-test-model.generator.ts
CarlosCortizasCT: I thought about it, but the template library (Mustache) preserves the templates formatting so it seemed an unneeded step to me.

generators/src/new-test-model/templates/model/builders.spec.tpl
CarlosCortizasCT: I'm all for not adding new libraries if not needed so let me take a look at lodash capabilities.

Perhaps this is a topic to seek for an agreement as I already found another templating library ([Squirrely](https://github.com/squirrellyjs/squirrelly)) used in the presets generation repository.

tdeekens: Squirrelly also looks nice in lean in terms of offering the minimal set of thing we need. Lodash also allows this inline JavaScript, looping and conditionals. If I'd had to choose I must admit Squirrelly looks more focussed a single task which I like.

CarlosCortizasCT: Ok, the I'll refactor the CLI to use Squirrely

CarlosCortizasCT: Migrated from Mustache to Squirrelly here: a2ff62e5806558737268dc4b6f868a933f0f92a8

generators/src/new-test-model/templates/package/package.json
```diff
@@ -0,0 +1,28 @@
...
+  "license": "MIT",
+  "publishConfig": {
+    "access": "public"
+  },
+  "main": "dist/commercetools-test-data-{{modelCodename}}.cjs.js",
+  "module": "dist/commercetools-test-data-{{modelCodename}}.esm.js",
+  "files": ["dist", "package.json", "LICENSE", "README.md"],
+  "dependencies": {
+    "@babel/runtime": "^7.17.9",
+    "@babel/runtime-corejs3": "^7.17.9",
```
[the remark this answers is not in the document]
CarlosCortizasCT: Updated here: a2ff62e5806558737268dc4b6f868a933f0f92a8

tdeekens (approved):
Thanks for making the changes. Appreciate taking the time 🤗.

PR #691, opened 2024-10-07
Update core helpers to support accurate typing for transformed models

nima-ct:
**Description**

This PR upgrades type accuracy for list builder helpers by introducing a second type parameter for the transformed output model (either GraphQL or REST).

- Previously, only the base model type was specified, representing the model from which the GraphQL or REST representations were derived. As a result, the returned data type did not accurately reflect the transformed output of these representations.

- Now, by specifying both the base model type and the transformed output type, the returned data type precisely matches the actual structure of the GraphQL and REST responses.

- For backward compatibility, the second type parameter will default to the base model (`Model`) if not provided.

- Users should now explicitly provide the `GraphqlModel` when needed:
```typescript
const graphqlList = buildGraphqlList<TDiscountCode, TDiscountCodeGraphql>(...)
```

### DIFF
--- a/.changeset/brown-spiders-impress.md
+++ b/.changeset/brown-spiders-impress.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-test-data/core': patch
+---
+
+Update core helpers to support accurate typing for transformed models
--- a/core/src/helpers.ts
+++ b/core/src/helpers.ts
@@ -115,17 +115,17 @@ const toGraphqlPaginatedQueryResult = <Model>(
   };
 };
 
-const buildField = <Model>(
+const buildField = <Model, TransformedModel = Model>(
   builder: Model | TBuilder<Model>,
   transformName: TTransformType = 'default',
   meta?: TBuildFieldMeta<Model>
-): Model => {
+): TransformedModel => {
   const buildName = convertTransformNameToBuildName(transformName);
   // @ts-ignore: TS does not know about the `Model` being an object.
   const builderField = builder?.[buildName];
   // We need to cast this to `() => Model` as otherwise the value is unknown.
   // We know it's a function because of the proxy builder.
-  const builderFn = builderField as (() => Model) | undefined;
+  const builderFn = builderField as (() => TransformedModel) | undefined;
   if (!builderFn) {
     throw new Error(
       `Builder with name '${buildName}' does not exist on field '${String(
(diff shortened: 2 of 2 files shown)


mario-priceless (approved):

nima-ct:
@emmenko @tdeekens any objections to these changes?

tdeekens (approved):
Thanks for the PR description and ping. This is a valuable improvement to how I can judge it.

nima-ct:
> I'm sorry I didn't have the time to add my review before this PR has been merged, but I think this change does not goes in the direction of all the work we're doing in the FE Chapter towards specialized builders instead of general ones which gets transformed.
> 
> I'm not opposed to this change per se, as it does not conflict with that work, but it won't be necessary in the future.
> 
> It would be nice for the FE Chapter to be pinged with this kind of changes because of all the deep changes we're working on ([reference](https://commercetools.atlassian.net/browse/FEC-85)).

Hey Carlos, that makes sense.

The changes came from a need I had while working on a new feature to resolve some type issues in one of my tests. However, if you believe these changes might cause issues or conflicts in the future, I’m open to reverting them. Let me know

CarlosCortizasCT:
> > I'm sorry I didn't have the time to add my review before this PR has been merged, but I think this change does not goes in the direction of all the work we're doing in the FE Chapter towards specialized builders instead of general ones which gets transformed.
> > I'm not opposed to this change per se, as it does not conflict with that work, but it won't be necessary in the future.
> > It would be nice for the FE Chapter to be pinged with this kind of changes because of all the deep changes we're working on ([reference](https://commercetools.atlassian.net/browse/FEC-85)).
> 
> Hey Carlos, that makes sense.
> 
> The changes came from a need I had while working on a new feature to resolve some type issues in one of my tests. However, if you believe these changes might cause issues or conflicts in the future, I’m open to reverting them. Let me know

My concern is about adding more code related to what are not the "legacy" patterns to build data models, but I think it's ok to keep your change.

Thanks.

tdeekens:
> I'm not opposed to this change per se, as it does not conflict with that work, but it won't be necessary in the future.

This is why I approved it.

> The changes came from a need I had while working on a new feature to resolve some type issues in one of my tests.

This is what I saw benefits in.

> t would be nice for the FE Chapter to be pinged with this kind of changes because of all the deep changes we're working on

Agreed let's all inform each other better.

PR #697, opened 2024-10-23
Setup a New Single Package

CarlosCortizasCT:
The goal of this PR is to publish a single package for all the models. We are currently publishing 40+ packages every time we release a new version of the test data models which is not good for the repository maintainability nor for its consumers.

The goal of this task is to set up a new package in this repository which aggregates all the test data models so we only need to release that one and consumers can simplify the way they use test data models and also how they stay up to date with the latest changes

**Note for reviewers**: all the new `package.json` files in the `presets/all-packages` subdirectories are autogenerated files (using `preconstruct`) and are needed to be able to consume each model independently.


presets/all-packages/associate-role/package.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "@commercetools/composable-commerce-test-data-associate-role",
```
bradd123, line 2: It requires a name to be able to import as a single model package.

generators/src/new-test-model/new-test-model.generator.ts
```diff
@@ -188,6 +195,61 @@ export const newTestModelGenerator: CodeGenerator = {
         });
     }
 
+    // 3. Update the single package preset
```
CarlosCortizasCT, line 198: We need to:
1. Create a file in the source folder where we export the new model
2. Update the `package.json` to make sure we have a new entry point for the new model, we include its source in the dist package and we add the internal dependency.
3. We create the "proxy" `package.json` for the independent entry point (using `preconstruct`)
4. We install the new internal dependency

models/product-type/src/attribute-localized-enum-value/types.ts
```diff
@@ -1,5 +1,5 @@
 import { AttributeLocalizedEnumValue } from '@commercetools/platform-sdk';
-import { TLocalizedStringGraphql } from '@commercetools-test-data/commons/src';
```
CarlosCortizasCT, line 2: Fixing an incorrect import.

models/product/src/product-data/types.ts
```diff
@@ -25,11 +24,7 @@ export type TProductDataRest = Omit<ProductData, 'categories'> & {
 };
 
 // This type only appears in the GraphQL representation
-export type TCategoryOrderHintGraphql = {
-  categoryId: keyof CategoryOrderHints;
-  orderHint: ValueOf<CategoryOrderHints>;
-  __typename: 'CategoryOrderHint';
-};
+export type TCategoryOrderHintGraphql = TCtpCategoryOrderHint;
```
CarlosCortizasCT, line 27: Use the generated GraphQL types is better.

models/type/src/custom-field-localized-enum-value/types.ts
```diff
@@ -1,5 +1,5 @@
 import { CustomFieldLocalizedEnumValue } from '@commercetools/platform-sdk';
-import { TLocalizedStringGraphql } from '@commercetools-test-data/commons/src';
+import { TLocalizedStringGraphql } from '@commercetools-test-data/commons';
```
CarlosCortizasCT, line 2: Fixing an incorrect import.

presets/all-packages/associate-role/package.json
```diff
@@ -0,0 +1,6 @@
+{
```
CarlosCortizasCT, line 1: All these `package.json` files are autogenerated by `preconstruct` and required for it to be able to export independent entry points.

presets/all-packages/src/attribute-group.ts
```diff
@@ -0,0 +1 @@
+export * from '@commercetools-test-data/attribute-group';
```
CarlosCortizasCT, line 1: This is how we re-export an internal package so it can be consumed independently.

presets/all-packages/src/index.ts
```diff
@@ -0,0 +1 @@
+export default {};
```
CarlosCortizasCT, line 1: We don't want to export everything at once. Consumers will need to import whatever they need independently.

presets/all-packages/README.md
```diff
@@ -0,0 +1,81 @@
+# @commercetools-test-data/cart
```
ahmehri, line 1: Is this a typo?

presets/all-packages/README.md
```diff
@@ -0,0 +1,81 @@
...
+
+This package provides access to all the data model used in the commercetools APIs .
```
ahmehri, line 3:
```suggestion
This package provides access to all the test data models used in the commercetools APIs .
```

presets/all-packages/README.md
CarlosCortizasCT: Yes.

Updated here: 9b894da0e0dde37d04c180a62347d228b9032a0b

presets/all-packages/README.md
CarlosCortizasCT: Updated here: 9b894da0e0dde37d04c180a62347d228b9032a0b

ahmehri (approved):

tdeekens (approved):

generators/src/new-test-model/new-test-model.generator.ts
```diff
@@ -188,6 +195,61 @@ export const newTestModelGenerator: CodeGenerator = {
...
+      // Generate the re-export proxy file
+      writeFileSync(
+        join(presetDirectoryPath, 'src', `${modelCodename}.ts`),
+        `export * from '@commercetools-test-data/${modelCodename}';`
+      );
+      // Update package.json
+      const presetPackageJson = (
+        await import(join(presetDirectoryPath, 'package.json'))
+      ).default;
+      presetPackageJson.files.push(modelCodename);
```
ismaelocaramelo, line 217: maybe worth to add some guards here if the structure isn't expected?

ismaelocaramelo (approved):
LTGM, I've just added a small doubt without having full context.

generators/src/new-test-model/new-test-model.generator.ts
CarlosCortizasCT: Not sure if I follow.
Are you worried about the structure of the `package.json` file import?
If that's the case, the file is managed by us and the `files` property is mandatory. Without it the preset would not work so I think it's safe to assume it will exists.

Please let me know if I misunderstood your comment 🙏

PR #700, opened 2024-10-25
fix: improve TModelFieldsConfig<TModel> type

ragafus:


CarlosCortizasCT:
Hi @ragafus ✋ 

I'm not sure about the background and goal of this PR but just wanted to let you know we have a [ticket](https://commercetools.atlassian.net/browse/FEC-62) in the FE chapter we want to use to improve typescript support for test data models, specially regarding the typing of fields configuration.

Our idea is to do some research on the topic and create an RFC where we document what are current status look like, what we think we can improve and how we can do it.

If you want to work on this topic, perhaps we can have some kind of initial alignment.

PR #703, opened 2024-10-31
[FEC-88] Migrating test data model Inventory Entry

jmcreasman:
As part of the new test data models implementation patterns, we not only will be using it for new data models but we also need to apply it for existing ones. That means we will need to eventually migrate all the current data models those new implementation patterns.

We already have [documentation](https://github.com/commercetools/test-data/blob/main/docs/guidelines/migrating-existing-model.md) about how this process looks like but we also need to have a reference example engineers can use to compare. That's what this PR is for using the InventoryEntry data model.


jmcreasman:
I wasn't sure if this PR was also supposed to delete the two unneeded files or if that would come later?

models/inventory-entry/src/fields-config.ts
```diff
@@ -0,0 +1,55 @@
...
+    ...commonFieldsConfig,
+  },
+};
+export const graphqlFieldsConfig: TModelFieldsConfig<TInventoryEntryGraphql> = {
+  fields: {
+    ...commonFieldsConfig,
+    __typename: 'InventoryEntry',
+    supplyChannelRef: fake((f) => f.string.uuid()),
+  },
+  postBuild: (model) => {
```
jmcreasman, line 44: Not 100% sure if I did this part correctly?

models/inventory-entry/src/types.ts
```diff
@@ -7,6 +7,9 @@ import type { TReferenceGraphql } from '@commercetools-test-data/commons';
 import type { TBuilder } from '@commercetools-test-data/core';
 
 // Default
+/**
+ * @deprecated use `TInventoryEntryRest` instead
+ */
 export type TInventoryEntry = Omit<InventoryEntry, 'supplyChannel'> & {
   supplyChannel: Channel;
 };
```
jmcreasman, line 15: For this PR do I need to do anything about the `InventoryEntryDraft`? I didn't see anything about that in the docs.

models/inventory-entry/src/index.ts
```diff
@@ -1,6 +1,29 @@
...
+
+export const InventoryEntryGraphql = {
+  random: GraphqlModelBuilder,
+  presets: modelPresets.graphqlPresets,
+};
+
+/**
+ * @deprecated Use `InventoryEntryRest` or `InventoryEntryGraphql` exported models instead of `InventoryEntry`.
+ */
+export const deprecatedInventoryEntry = {
```
jmcreasman, line 26: Originally I had this looking like the doc example but I was getting this error when trying to commit:

models/inventory-entry/src/index.ts:8:1
  ✖   8:1   Multiple exports of name random.   import/export
  ✖   8:1   Multiple exports of name presets.  import/export
  ✖  26:14  Multiple exports of name random.   import/export
  ✖  27:14  Multiple exports of name presets.  import/export
  

So I changed it to this and that made it happier. Let me know if that's not correct though.

models/inventory-entry/src/fields-config.ts
CarlosCortizasCT: Looks good to me 👍

models/inventory-entry/src/presets/index.ts
```diff
@@ -1,3 +1,29 @@
-const presets = {};
+import type { TBuilder } from '@commercetools-test-data/core';
```
CarlosCortizasCT, line 1: The index file of presets should export just plain objects with the presets for each representation.
In this case, we don't have any presets implemented for this model, so it would be something like this:
```ts
```

.changeset/metal-mice-sit.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-test-data/inventory-entry': minor
+---
+
+feat(inventory entry): update test data model migrating to new implementation
```
CarlosCortizasCT, line 5:
```suggestion
Refactored `InventoryEntry` and `InventoryEntryDraft` models to use the new implementation patterns.
This means the existing models are now deprecated (though we still support them for backwards compatibility) and consumers should move to the new ones implemented:
* `InventoryEntryRest`
* `InventoryEntryGraphql`
* `InventoryEntryDraftRest`
* `InventoryEntryDraftGraphql`
```

models/inventory-entry/src/index.ts
```diff
@@ -1,6 +1,27 @@
+import {
+  RestModelBuilder,
+  GraphqlModelBuilder,
+  CompatInventoryEntryModelBuilder,
+} from './builders';
+import * as modelPresets from './presets';
+
 export * as InventoryEntryDraft from './inventory-entry-draft';
```
CarlosCortizasCT, line 8: Please remember to also migrated the draft model.

models/inventory-entry/src/types.ts
CarlosCortizasCT: Yes, we need to migrate that model as well.

models/inventory-entry/src/index.ts
CarlosCortizasCT: I think that was because the deraft model has not been migrated yet.

models/inventory-entry/src/inventory-entry-draft/index.ts
```diff
@@ -1,2 +1,27 @@
-export { default as random } from './builder';
-export { default as presets } from './presets';
+// export { default as random } from './builder';
+// export { default as presets } from './presets';
```
CarlosCortizasCT, line 2: We can remove this commented code

models/inventory-entry/src/index.ts
CarlosCortizasCT, line 8: We need to export the three models with heir own names:
```suggestion
export * from './inventory-entry-draft';
```

CarlosCortizasCT:
@jmcreasman  We still need to refactor the sample data presets. I think it would be helpful to have a meeting to clarify this point.

jmcreasman:
> @jmcreasman We still need to refactor the sample data presets. I think it would be helpful to have a meeting to clarify this point.

Yes I generated it from the test data generation repo and added it in, but then realized the data it generated wasn't in the new format/implementation even when I made it `v2` like in the guide. I tested other entities like Channel and it worked as expected so I've asked the First Contact team about this and should have a reply this morning (many of them were out yesterday).

PR #709, opened 2024-11-04
Channel and LocalizedString adjustments

CarlosCortizasCT:
When trying to use last version of this repo packages in the MC monorepo we found out:
* It is better for compatibility models to have GraphQL as the default presets
* `Channel` model should have the `geoLocation` fileld populated by default
* `LocalizedString` helper was returning `undefined` in some situation where it should return `null`


models/attribute-group/src/types.ts
```diff
@@ -10,7 +10,7 @@ export type TAttributeGroupGraphql = Omit<
 > & {
   name: string;
   nameAllLocales?: TLocalizedStringGraphql | null;
-  description?: string;
```
CarlosCortizasCT, line 13: Many consumers GraphQL queries are expecting `null` instead of undefined so I'm extending these types.

mario-priceless (approved):

PR #713, opened 2024-11-12
FCT-1276-Price keys

lidiyacommercetools:
Added keys to prices and added additional prices to variants that use a distribution channel.

Also deleted a corrupt product image for double walled espresso glass.

### DIFF
--- a/.changeset/few-ghosts-count.md
+++ b/.changeset/few-ghosts-count.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-test-data/product': minor
+---
+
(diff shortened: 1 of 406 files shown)


stephsprinkle (approved):
Confirmed that all the variants that have price drafts with channels now have duplicate prices without channels.

ByronDWall (approved):

tylermorrisford (approved):

CarlosCortizasCT:
I'd like to release a new version of the packages and I noticed this PR is approved but not merged.

Is it ok to merge so it gets published with the next upcoming release?

PR #721, opened 2024-11-26
FCT-1240: Add product variant keys

lidiyacommercetools:
Added keys to product variants for the B2C data set. Update is needed to support import & export functionality in the MC.


tylermorrisford (approved):
Spot check looks good!

ByronDWall (approved):
spot check looks good, thanks for adding all this!

PR #722, opened 2024-11-29
Update GraphQL types package setup to include all types available

CarlosCortizasCT:
## Goal
The main goal of this PR is to improve the way we have to define the GraphQL Typescript types for the test data models.

## Description
One of the first steps we need to take while building a new data model is to define the Typescript types we will use to ensure the model object instances structure.

We currently support two representations of a data model: a REST one and a GraphQL one.

Defining the type of the test data model REST representation is easy because we can rely on the `@commercetools/platform-sdk` package which already contains them all we can usually just defined the REST test data model main type like this:
```ts
export type TChannelRest = Channel;
```

However, when defining the type for the GraphQL representation things were not that easy as we didn't have a similar package so we were just using the REST type as a reference and adjust it like this:
```ts
type TChannelGraphql = Omit<
  Channel,
  // In GraphQL, these properties have different shapes.
  'name' | 'description' | 'createdBy' | 'lastModifiedBy' | 'geoLocation'
> & {
  __typename: 'Channel';
  createdBy?: TClientLoggingGraphql;
  lastModifiedBy?: TClientLoggingGraphql;
  name?: string | null;
  description?: string | null;
  descriptionAllLocales?: TLocalizedStringGraphql | null;
  geoLocation?: TGeometryGraphql;
};
```

We already have an internal package in this repository (`@commercetools-test-data/graphql-types`) which is sued to hold the GraphQL types for some of the backend services we need to create test data models for, but it was only exporting types for the MC Settings service.

In this PR we are updating that package to export all the GraphQL Typescript types for all the service we need to we can use it the same way we use the `@commercetools/platform-sdk` pacakage for the REST types.

After the setup is updated, we can define the main types of a test data model like this:
```ts
import type { Channel } from '@commercetools/platform-sdk';
import type { TCtpChannel } from '@commercetools-test-data/graphql-types';

export type TChannelRest = Channel;
export type TChannelGraphql = TCtpChannel;
```

In this PR we're also updating the [test data models generation CLI](https://github.com/commercetools/test-data/tree/main/generators) to take this into account and do it automatically.

There are some caveats that had to be addressed though.
They are explained in inline comments in this PR.

There's also a follow-up [PR](https://github.com/commercetools/test-data/pull/724) to update docs accordingly.


codegen.ctp.yml
```diff
@@ -0,0 +1,23 @@
+schema:
```
CarlosCortizasCT, line 1: We were missing the types for this service.

generators/src/new-test-model/new-test-model.generator.ts
```diff
@@ -24,6 +31,22 @@ export const newTestModelGenerator: CodeGenerator = {
     });
     const modelCodename = snakeCase(modelName).replaceAll('_', '-');
 
+    const { modelOwningService } = await prompts({
```
CarlosCortizasCT, line 34: Since the GraphQL types from the services are prefixed, we need to ask about which service provide the type the developer wants to generate in order to correctly build the `import` statements in the templates.

graphql-types/src/types-mappings.json
```diff
@@ -0,0 +1,8 @@
+{
```
CarlosCortizasCT, line 1: This is the other reason why we use a post-processor for the GraphQL types generation.

We have instances where the GraphQL schema defines a field with an abstract type, but the actual service returns an implementation.

By default, the generated Typescript types will use the abstract type, which won't allow consumers to use the actual data returned by the service.

Here I just implemented a basic example.
The `Geometry` types in the GraphQL schema looks like this:
```graphql
type Geometry {
  type: String!
}
```

but it has only one implementation which is the `Point`:
```graphql
type Geometry {
  type: String!
  coordinates: [Float!]!
}
```

With this configuration we can adjust the generated types to use the implementation type instead of the abstract one.

There are other use cases where an abstract type might have several implementations (`Money` has `HighPrecissionMoney` and `CentPrecissionMoney`). In those cases we are already using just one of the implementations in the test data models, so it wouldn't be much different.

We could even think on extending this first solution so we can decide the swapping per model. Something like this:
```json
  "ctp": [
    {
      "from": "ModelA.fieldA",
      "to": "HightPrecissionMoney"
    },
    {
      "from": "ModelB.fieldA",
      "to": "CentPrecissionMoney"
    }
  ]
}
```

Anyway, let's take it step by step.

graphql-types/src/types-post-processor.js
```diff
@@ -0,0 +1,60 @@
+const fs = require('fs');
```
CarlosCortizasCT, line 1: I know this is a simplistic implementation, but I think it's better to start simple and only add complexity when we realize we need it.

nima-ct (approved):
LGTM and great improvement 👍

generators/src/new-test-model/new-test-model.generator.ts
```diff
@@ -24,6 +31,22 @@ export const newTestModelGenerator: CodeGenerator = {
...
+      type: 'select',
+      name: 'modelOwningService',
+      message: 'What is the name of the service that owns this model',
+      choices: [
+        { title: 'Core (Organization related models)', value: 'core' },
+        { title: 'CTP (Project related models)', value: 'ctp' },
+        { title: 'MC Gateway', value: 'mc' },
+        { title: 'MC Settings', value: 'settings' },
```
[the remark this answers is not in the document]
CarlosCortizasCT: I understand your point but my concern is that users will have the same hesitation I had related to the difference between CTP and Core.
If we use your suggestions, perhaps engineers will still hesitate about what "CoCo Core API" is for.

graphql-types/src/types-mappings.json
CarlosCortizasCT: Seems to work like charm.

Updated here: 9c903f870743350ce9cb61d227fb7e65c53c39b0

graphql-types/src/types-post-processor.js
CarlosCortizasCT: Updated here: https://github.com/commercetools/test-data/commit/9c903f870743350ce9cb61d227fb7e65c53c39b0

codegen.core.yml
```diff
@@ -19,3 +19,4 @@ extensions:
     hooks:
       afterAllFileWrite:
         - prettier --write
+        - node ./graphql-types/src/types-post-processor.js --types-prefix=TCore
```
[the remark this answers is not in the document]
CarlosCortizasCT: The post-processor has been updated and simplified after I started using the `useImplementingTypes` configuration property you suggested.

graphql-types/src/types-mappings.json
emmenko: Do we still need this file if the `useImplementingTypes` option works?

graphql-types/package.json
```diff
@@ -15,5 +15,10 @@
   },
   "main": "dist/commercetools-test-data-graphql-types.cjs.js",
   "module": "dist/commercetools-test-data-graphql-types.esm.js",
-  "files": ["dist", "package.json", "LICENSE", "README.md"]
+  "files": ["dist", "package.json", "LICENSE", "README.md"],
+  "dependencies": {
+    "@babel/runtime-corejs3": "^7.21.0",
+    "@graphql-codegen/plugin-helpers": "5.1.0",
+    "lodash": "4.17.21"
+  }
```
emmenko, line 23: Are these dependencies still needed?

emmenko (approved):
Nice work, this is very helpful!

graphql-types/package.json
CarlosCortizasCT: No they are not.

Removed them in this commit: 2b5f91166820a4c7a6282aedf7455403b9a73937

PR #724, opened 2024-12-02
Update data models guidelines

CarlosCortizasCT:
This is a follow-up of #722 where we set up GraphQL types generation so we don't need to define them in each test data model.

We're updating the docs to reflect those changes in this PR.

Also, we gathered some initial feedback around creating new models and the concept of referenced models does not seem to be very clear so I added a new section trying to explain a bit what they are and how we handle them.


CarlosCortizasCT:
We can't merge this PR until the previous one (#722) is merged first.

obulaworld (approved):

PR #726, opened 2024-12-09
[FEC-49] ShippingInfo model

CarlosCortizasCT:
In this PR we're introducing the `ShippingInfo` model to the `cart` package.
It required to also include other cart child models.


ddouglasz (approved):
Thank you 🙏🏾

nima-ct (approved):

PR #729, opened 2024-12-18
feat(b2b): updateing shipping methods for b2b sample data

NickDevG:
## Summary

Update B2B sample data shipping methods information according to CoFE request.

Screenshot 2024-12-18 at 12 43 50


antoniolodias (approved):
Code looks good! 
I am still a bit confused about having specific presets on a public repo though

CarlosCortizasCT:
> Code looks good! I am still a bit confused about having specific presets on a public repo though

@antoniolodias That's a very valid question.

In this case, all those presets are part of the `sampleData` dataset that is used as part of the new projection creation use flow when they want to populate the new project with some data.

The FE chapter already had the same thought as you and proposed to move all the `sampleData` related stuff out of it ([reference](https://commercetools.atlassian.net/browse/FEC-158)).
The proposal has been approved and now it's just a matter of time for that to happen.

As part of that effort, the chapter will be including some guidelines in this repository about what kind of presets are allowed to exist here.

PR #631, opened 2024-08-14
New `ProductProjection` model

CarlosCortizasCT:
## Description

As part of the effort to [remove the local test data models](https://commercetools.atlassian.net/browse/FEC-26) we have in the Merchant Center monorepo, we noticed the `ProductProjection` model did not exist yet in this repository, so this PR introduces it.

[Here](https://docs.commercetools.com/api/projects/productProjections) is the public technical documentation about this model.

PR #738, opened 2025-01-08
Migrate ProductProjection model

CarlosCortizasCT:
The `ProductProjection` model was one of the first ones [we created](https://github.com/commercetools/test-data/pull/631) when we started the initiative to remove local test data models from the MC frontend monorepo but we never got to actually use it.

Now that I was reviewing the pending work we have, I noticed we had this PR that was trying to address some issues related to this model but that was something we tried before the introduction of the new implementation patterns so I took the opportunity to start cleaning up the model and migrate it to the new patterns.

The main compatibility issue for this model to be usable was it had an incorrect preset name.


models/commons/src/reference/transformers.ts
```diff
@@ -1,14 +1,23 @@
...
-      obj: omit(fields, ['typeId']),
-    }),
+    buildFields: ['obj'],
+    replaceFields: ({ fields }) => {
+      return {
+        ...fields,
+        obj:
+          (fields.obj as TExpandedReferenceObject) || omit(fields, ['typeId']),
+      };
+    },
```
CarlosCortizasCT, line 20: This is so we can forward to the `obj` object all the properties we potentially have received so, in case the model was populated with a reference object, the model instance has its properties.

models/product-projection/src/builder.spec.ts
```diff
@@ -27,99 +25,9 @@ describe('builder', () => {
...
-        lastModifiedAt: expect.any(String),
-        reviewRatingStatistics: null,
-        priceMode: null,
-      })
-    )
-  );
-
-  it(
-    ...createBuilderSpec<TProductProjection, TProductProjectionRest>(
+    ...createBuilderSpec<TProductProjectionRest, TProductProjectionRest>(
```
CarlosCortizasCT, line 28: Removed the default type test as we don't have it anymore.

models/product-projection/src/builders.ts
```diff
@@ -0,0 +1,25 @@
+import { createSpecializedBuilder } from '@commercetools-test-data/core';
```
CarlosCortizasCT, line 1: Part of the migration of this model to use the new implementation patterns.

models/product-projection/src/fields-config.ts
```diff
@@ -0,0 +1,131 @@
+import { LocalizedString, Reference } from '@commercetools-test-data/commons';
```
CarlosCortizasCT, line 1: Part of the migration of this model to use the new implementation patterns.

models/product-projection/src/index.ts
```diff
@@ -1,4 +1,14 @@
-export { default as random } from './builder';
-export { default as presets } from './presets';
+import * as modelBuilders from './builders';
```
CarlosCortizasCT, line 1: Since this model was not used yet, I'm not including the compat builder.

models/product-projection/src/presets/happy-cow-milk.ts
```diff
@@ -0,0 +1,52 @@
+import { LocalizedString, Reference } from '@commercetools-test-data/commons';
```
CarlosCortizasCT, line 1: Part of the migration of this model to use the new implementation patterns.

models/product-projection/src/types.ts
```diff
@@ -1,73 +1,13 @@
...
-} from '@commercetools/platform-sdk';
-import {
-  TLocalizedStringGraphql,
-  TReferenceGraphql,
-} from '@commercetools-test-data/commons';
+import { ProductProjection } from '@commercetools/platform-sdk';
 import type { TBuilder } from '@commercetools-test-data/core';
-import { TState } from '@commercetools-test-data/state';
-import { TTaxCategory } from '@commercetools-test-data/tax-category';
+import type { TCtpProductProjection } from '@commercetools-test-data/graphql-types';
```
CarlosCortizasCT, line 3: Simplified the types by leveraging the generated GraphQL ones.

kark (approved):
Thanks, looks good to me 👍
We need to add the changeset, right?

ByronDWall (approved):

CarlosCortizasCT:
> Thanks, looks good to me 👍 We need to add the changeset, right?

Thanks for the reminder, Kacper 🙇 

Changesets added here: bfa97b06d450151e9f24ebe2e5d9dc86cef85f42

PR #741, opened 2025-01-15
Migrate Reference model

CarlosCortizasCT:
The `Reference` model had one main downside since it was not actually possible to populate the `obj` property of its REST representation.
We added an embarrassing walk-around in [this PR](https://github.com/commercetools/test-data/pull/738/files#diff-31d0b3cd9b7649d66e89bb0f86f0d3f5aada2d6ad20d846b26b7eb047e284459) that we wanted to address as a follow-up so we're migrating the `Reference` model to the new implementation patterns which solves the issue and improves its maintainability.

**NOTE**: There are many files involved in this PR but it's because I took the time to simplify the presets and removed many files (no need to maintain barrel files for this local files and so many similar tests). No need to focus on that part as we're keeping the same API.


models/commons/src/reference/presets/_shared/reference-presets-builder.spec.ts
```diff
@@ -0,0 +1,50 @@
+import { buildReferencePresets } from './reference-presets-builder';
+
+const presets = buildReferencePresets('associate-role');
+
+describe('AssociateRoleReference preset', () => {
```
ByronDWall, line 5: I assume this test is a stand in for the tests for each reference preset type?  Which is fine, and I'm fine with only testing one preset as a stand in for all of them.

but if the test ends up needing to be more comprehensive for whatever reason it would be pretty straightforward to loop over every reference type in `models/commons/src/reference/constants.ts` using jests [test.each](https://jestjs.io/docs/api#testeachtablename-fn-timeout)

ByronDWall (approved):
Overall this looks good to me, and it definitely removes a lot of boilerplate, which is good to see.

models/commons/src/reference/index.ts
```diff
@@ -1,5 +1,30 @@
...
+  presets: modelPresets.restPresets,
+};
+
+export const ReferenceGraphql = {
+  random: GraphqlModelBuilder,
+  presets: modelPresets.graphqlPresets,
+};
+
+/**
+ * @deprecated Use `ReferenceRest` or `ReferenceGraphql` exported models instead of `Reference`.
```
tdeekens, line 24: This is nice!

models/commons/src/reference/reference-draft/builders.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+    validateModel(graphqlModel);
+  });
+
+  it('builds a populated GraphQL model', () => {
+    const graphqlModel = ReferenceDraftGraphql.random()
+      .id('12345')
+      .typeId('foo')
+      .build();
+
+    validateModel(graphqlModel, {
```
tdeekens, line 53: Just a thought: we should also extend our Jest/Vitest matchers. So in the future we could have `expect(graphqlModel).toMatchModelContainingFields()` or a different matcher name.

tdeekens (approved):
Really nice work. Thanks for being so persistent on this!

models/commons/src/reference/reference-draft/builders.spec.ts
CarlosCortizasCT: To be honest, I don't see too much value on that as consumers using that function will need to provide very much the same values that are currently used to validate a model.
I find this local helpers a little more helpful, specially in migrated models, as we have a specific set of properties to match in different local tests.

Happy to learn more about your idea in a sync if you want to.

models/commons/src/reference/presets/_shared/reference-presets-builder.spec.ts
CarlosCortizasCT: Thanks for the suggestion.

My reasoning for having just one test file is that we're not testing the presets but the underlying helper (`reference-presets-builder`) those presets use.
Testing each presets does not seem very valuable to me as the only think it changes is the name of the `typeId` and we're already using a constant for populating that property.

models/commons/src/reference/reference-draft/builders.spec.ts
tdeekens: We can take it to our sync.

PR #743, opened 2025-01-17
chore: migrate cart model

YahiaElTai:
This PR migrates the `cart` model to the new structure while maintaining the compatibility layer.


models/cart/src/cart/index.ts
```diff
@@ -1,8 +1,29 @@
...
+
+export const CartGraphql = {
+  random: GraphqlModelBuilder,
+  presets: modelPresets.default,
+};
+
+/**
+ * @deprecated Use `CartRest` or `CartGraphql` exported models instead of `Cart`.
+ */
+
```
CarlosCortizasCT, line 27: **nit**: we need to remove the extra empty line for the JsDoc to actually be used.

models/cart/src/cart/builder.ts
```diff
@@ -1,12 +1,32 @@
-import { Builder } from '@commercetools-test-data/core';
-import generator from './generator';
-import transformers from './transformers';
-import type { TCart, TCreateCartBuilder } from './types';
+import {
```
CarlosCortizasCT, line 1: **nit**: this file and the spec one should be renamed to be plural: builder -> builders

models/cart/src/cart/fields-config.ts
```diff
@@ -0,0 +1,91 @@
...
+  shippingMode,
+  taxCalculationMode,
+  taxMode,
+  taxRoundingMode,
+} from './constants';
+import { TCartGraphql, TCartRest } from './types';
+
+const [getOlderDate, getNewerDate] = createRelatedDates();
+
+const commonFieldsConfig = {
```
CarlosCortizasCT, line 28: Bear in mind the nested entities are handled differently in REST and GraphQL.
You can take a look at [this section](https://github.com/commercetools/test-data/blob/main/docs/guidelines/creating-new-model.md#handling-references) of the new models docs for details.

models/cart/src/cart/builder.spec.ts
```diff
@@ -1,267 +1,199 @@
...
+        clientId: expect.any(String),
+        customerRef: expect.objectContaining({ typeId: 'customer' }),
+        userRef: expect.objectContaining({ typeId: 'user' }),
+        __typename: 'Initiator',
+      }),
+    })
   );
+};
+
+describe('Cart model builders', () => {
```
CarlosCortizasCT, line 183: We should also test the legacy builder.

CarlosCortizasCT (commented):
I left some comments as a first review.

models/cart/src/cart/fields-config.ts
YahiaElTai: I will look into this now! Thanks

models/cart/src/cart/builder.spec.ts
YahiaElTai: Done

models/cart/src/cart/index.ts
YahiaElTai: Done

models/cart/src/cart/fields-config.ts
YahiaElTai: okay i took a look at the docs and this is what I understood, please correct me if this is wrong, cc @CarlosCortizasCT 

**Rest**: has field called store which is actually a reference to a store

**Graphql has 2 fields**
- store which is the actual store
- storeRef which is a reference to a store

YahiaElTai: Thanks for all the details. I have pushed a commit with the new changes to the cart model, please let me know if everything is correct so I can continue with the tests.

The `obj` property is handled at the Reference model level with the transformers 
and I have made all the needed changes for postBuild to ensure that they refs are linked properly 
cc @CarlosCortizasCT

models/cart/src/cart/builder.ts
CarlosCortizasCT: Please remember to also rename the test file.

models/cart/src/cart/builder.spec.ts
```diff
@@ -1,267 +1,250 @@
...
+      CartGraphql.random().lineItems([LineItem.random()])
+    ).build();
+
+    validateGraphqlModel(graphqlModel);
+  });
+});
+
+describe('Cart model compatibility builders', () => {
+  it('builds a REST model', () => {
+    const restModel = populateRestModel(Cart.random()).buildRest();
```
CarlosCortizasCT, line 238: Should we be populating the `lineItems` here as well to have all tests with the same set of data?

models/cart/src/cart/fields-config.ts
YahiaElTai: I have updated also the cart draft tests and I think now I'm ready for a full review.

Of course the only thing missing now is the sample data which I wasn't able to generate from test-data-generation model, I have left you a slack message about that for whenever you have a moment.

models/cart/src/cart/fields-config.ts
```diff
@@ -0,0 +1,130 @@
...
+    refusedGifts: fake(() => [Reference.random().typeId('cart-discount')]),
+  },
+};
+
+export const graphqlFieldsConfig: TModelFieldsConfig<TCartGraphql> = {
+  fields: {
+    ...commonFieldsConfig,
+    __typename: 'Cart',
+    customerGroup: fake(() => CustomerGroup.random()),
+    customerGroupRef: fake(() => Reference.random().typeId('customer-group')),
```
CarlosCortizasCT, line 90: No need to populate values here as it will be done in the `postBuild` hook.
We can leave them as `null`.

models/cart/src/cart/builder.ts
YahiaElTai: Done! sorry 😅

YahiaElTai:
I migrated all the presets and their spec files.
`sample-data-fashion/` presets I migrated them manually since they are not part of the test-data-generation script.

models/cart/src/index.ts
```diff
@@ -5,7 +5,7 @@ export * from './line-item/types';
 
 // Export models
 export * as Cart from './cart';
-export * as CartDraft from './cart/cart-draft';
+export * from './cart/cart-draft';
```
YahiaElTai, line 8: I changed this to maintain the same imports for consumers. since the cart package exports main different things and has an extra `index.ts` file, I needed to change this export to maintain the same behavior for consumers

models/cart/src/cart/fields-config.ts
```diff
@@ -0,0 +1,130 @@
...
+    customerGroup: fake(() => CustomerGroup.random()),
+    customerGroupRef: null,
+    businessUnit: fake(() => Company.random()),
+    businessUnitRef: null,
+    store: fake(() => Store.random()),
+    storeRef: null,
+    refusedGifts: fake(() => [CartDiscount.random()]),
+    refusedGiftsRefs: null,
+    customer: null,
+    placement: null,
```
CarlosCortizasCT, line 98: Not sure why we have these two properties here but not in the REST representation.
Do they actually only exist in the GraphQL representation?

models/cart/src/index.ts
CarlosCortizasCT: I'm not sure I follow. 
Would it be ok to have a quick sync about this? 🙏

YahiaElTai: sure I'm available for a sync anytime that is good for you

models/cart/src/cart/fields-config.ts
YahiaElTai: as far as the CTP generated types yes, these were required (typescript indicated only in the rest representation)

CarlosCortizasCT (approved):
Good to go from my side 🚀 

Thanks a lot for the efforts, Yahia 🙇

PR #744, opened 2025-01-20
PRC-3862: Add new models for CartDiscountPatternTarget, CountOnLineItemUnits, CountOnCustomLineItemUnits

nima-ct:
This PR introduces new models for the CartDiscountPatternTarget, CountOnLineItemUnits, and CountOnCustomLineItemUnits in the `@commercetools-test-data/cart-discount` package, using the our new approach.


mario-priceless (approved):
Great! 
note: Do we have a follow-up task in place to implement this new test model into our codebase? If not, let's make sure we add one.

dogayuksel (approved):

nima-ct:
> Great! note: Do we have a follow-up task in place to implement this new test model into our codebase? If not, let's make sure we add one.

Yes, here https://commercetools.atlassian.net/browse/PRC-3790

PR #751, opened 2025-02-11
refactor: move DiscountedPrice models from commons to product-discount

ragafus:
`DiscountedPriceRest`, `DiscountedPriceGraphql`, `DiscountedPriceDraftRest` and `DiscountedPriceDraftGraphql` entities existed in `@commercetools-test-data/commons`, although they were not exported.

While exporting these entities I realized that they were introducing a circular dependency that was making tests fail.

To break that circular dependency the entities are moved to `@commercetools-test-data/product-discount` and exported.


CarlosCortizasCT (approved):

mario-priceless (approved):

nima-ct (approved):

tdeekens (approved):
I agree I think in this case it doesn't make sense to locate it under commons and should be moved outside.

PR #764, opened 2025-03-20
chore(business-unit): add inheritedStores field to Company and Division models

YahiaElTai:
This PR adds missing inheritedStores field to the generator of Company and Division according to the new SDK (in the default and rest and graphql transformers)

### DIFF
--- a/models/business-unit/src/company/generator.ts
+++ b/models/business-unit/src/company/generator.ts
@@ -30,6 +30,7 @@ const generator = Generator<TCompany>({
     key: fake((f) => f.lorem.slug(2)),
     status: oneOf(...Object.values(status)),
     stores: [],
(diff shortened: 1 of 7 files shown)


leventekobor (approved):

models/business-unit/src/company/generator.ts
```diff
@@ -30,6 +30,7 @@ const generator = Generator<TCompany>({
     key: fake((f) => f.lorem.slug(2)),
     status: oneOf(...Object.values(status)),
     stores: [],
+    inheritedStores: [],
```
YahiaElTai, line 33: This was missing so the model didn't include this field even though the SDK has it.  the presence of this field is conditional which is the problem but including it here will save us some headache when testing

PR #766, opened 2025-03-26
Prc 3957 update the cart discount test data model to reflect the removal of exclude

mario-priceless:
excludeCount property has been deprecated and should be removed from all our cart discount related models

### DIFF
--- a/.changeset/clever-mayflies-joke.md
+++ b/.changeset/clever-mayflies-joke.md
@@ -1,5 +1,0 @@
----
-'@commercetools-test-data/cart-discount': minor
----
-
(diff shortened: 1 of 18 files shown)


.changeset/clever-mayflies-joke.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-test-data/cart-discount': minor
+---
+
+excludeCount property from cart discounts models
```
dogayuksel, line 5: remove ... ?

ragafus (approved):

PR #769, opened 2025-03-31
Migrate CustomLineItem to new model

krishhna123:
This PR is to migrate the `CustomLineItem` and `CustomLineItemDraft` to a new model with the help of [this guideline](https://github.com/commercetools/test-data/blob/main/docs/guidelines/migrating-existing-model.md)


models/cart/src/custom-line-item/field-config.ts
```diff
@@ -0,0 +1,63 @@
...
+  type TModelFieldsConfig,
+  oneOf,
+} from '@commercetools-test-data/core';
+import { customLineItemPriceMode } from '../cart/constants';
+import { TCustomLineItemRest, TCustomLineItemGraphql } from './types';
+
+const commonFieldsConfig = {
+  id: fake((f) => f.string.uuid()),
+  key: fake((f) => f.string.uuid()),
+  name: null,
```
CarlosCortizasCT, line 17: One of the patterns I think we should follow is to populate in the default model all its required fields. Taking a look at the REST type, I can see these ones marked as required: `id`, `name`, `money`, `taxedPricePortions`, `totalPrice`, `slug`, `quantity`, `state`, `perMethodTaxRate`, `discountedPricePerQuantity` and `priceMode`.

So I'm missing `name`, `totalPrice`.
Also, bear in mind the `name` property has a different shape between the REST and GraphQL models.

models/cart/src/custom-line-item/custom-line-item-draft/field-config.ts
```diff
@@ -0,0 +1,40 @@
...
+import {
+  TCustomLineItemDraftRest,
+  TCustomLineItemDraftGraphql,
+} from '../types';
+
+const commonFieldsConfig = {
+  key: fake((f) => f.string.uuid()),
+  name: fake(() => LocalizedStringDraft.random()),
+  quantity: fake((f) => f.number.int(9)),
+  money: fake(() => MoneyDraft.random()),
```
CarlosCortizasCT, line 20: It seems the types for the `money` prop are not the same for REST and GRAPHQL models.

models/cart/src/custom-line-item/index.ts
```diff
@@ -1,5 +1,24 @@
-export * as Cart from '.';
+import {
```
CarlosCortizasCT, line 1: With this new way of exporting the models we also need to update the `index.ts` file of the package. 
If we keep it like this, consumers using the `CustomLineItem` model would need to update their imports to use `CustomLineItem.CustomeLineItem` instead.

This is also true for the draft model.

models/cart/src/custom-line-item/custom-line-item-draft/presets/change-history-data/with-usd-currency-code.spec.ts
```diff
@@ -1,17 +1,37 @@
...
-      customLineItemWithUsdCurrencyCode().build<TCustomLineItemDraft>();
+  it('[REST] should return a custom line item with the money currencyCode set to `USD`', () => {
+    validateMoneyField(restPreset().build());
+  });
+
+  it('[GraphQL] should return a custom line item with the money currencyCode set to `USD`', () => {
+    validateMoneyField(graphqlPreset().build());
+  });
+
+  it('[Compat - REST] should return a custom line item with the money currencyCode set to `USD`', () => {
```
CarlosCortizasCT, line 30: I would also add a test where we use the compatibility builder but we call the `build` function (it should produce a result equal to the `rest` one).

models/cart/src/custom-line-item/builders.spec.ts
```diff
@@ -0,0 +1,121 @@
...
+
+  it('builds a GraphQL model', () => {
+    const graphqlModel = populateGraphqlFields(
+      CustomLineItemGraphql.random()
+    ).buildGraphql();
+    validateGraphqlFields(graphqlModel);
+  });
+});
+
+describe('CustomLineItem compatibility builder', () => {
```
CarlosCortizasCT, line 109: I would also add a test where we use the compatibility builder but we call the build function (it should produce a result equal to the rest one).

models/cart/src/custom-line-item/custom-line-item-draft/builders.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+  it('builds a GraphQL model', () => {
+    const customLineItemDraftGraphql = CustomLineItemDraftGraphql.random()
+      .custom(CustomFieldBooleanType.random())
+      .buildGraphql();
+
+    validateGraphqlFields(customLineItemDraftGraphql);
+  });
+});
+
+describe('CustomLineItemDraft compatibility builders', () => {
```
CarlosCortizasCT, line 88: I would also add a test where we use the compatibility builder but we call the build function (it should produce a result equal to the rest one).

CarlosCortizasCT (commented):
Thanks for working on this migration, Krishna 🙇 

I left some comments with some small things I believe we should take into consideration.

I also wanted to know your point of view regarding the guidelines we have in the docs folder. Do you think they're useful enough? Are you missing something on them?

models/cart/src/custom-line-item/custom-line-item-draft/presets/change-history-data/with-usd-currency-code.spec.ts
krishhna123: Added test for `build` function on compatibility builder

models/cart/src/custom-line-item/types.ts
```diff
@@ -3,18 +3,35 @@ import {
...
+    | TCustomLineItemDraftRest
+    | TCustomLineItemDraftGraphql,
+> = () => TBuilder<TCustomLineItemModel>;
 
-//scaffolding only
-export type TCustomLineItemGraphql = TCustomLineItem & {
-  __typename: 'CustomLineItem';
-};
-export type TCustomLineItemDraftGraphql = TCustomLineItemDraft;
+export type TCreateCustomLineItemDraftBuilder<
```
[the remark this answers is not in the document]
krishhna123: Removed `TCreateCustomLineItemDraftBuilder` and using `TCreateCustomLineItemBuilder` for both models

models/cart/src/custom-line-item/builders.spec.ts
```diff
@@ -0,0 +1,121 @@
...
+  TCustomLineItemRest,
+  TCustomLineItemGraphql,
+} from './types';
+import {
+  CustomLineItemRest,
+  CustomLineItemGraphql,
+  CustomLineItem,
+} from './index';
+
+const populateCommonFields = <
```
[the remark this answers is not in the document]
krishhna123: Done

models/cart/src/custom-line-item/builders.spec.ts
```diff
@@ -0,0 +1,121 @@
...
+  TModel extends TCustomLineItem | TCustomLineItemGraphql | TCustomLineItemRest,
+>(
+  model: TBuilder<TModel>
+) => model.money(Money.random()).custom(CustomFieldBooleanType.random());
+
+const populateGraphqlFields = (model: TBuilder<TCustomLineItemGraphql>) =>
+  populateCommonFields(model)
+    .nameAllLocales(LocalizedString.random())
+    .taxCategory(TaxCategory.random());
```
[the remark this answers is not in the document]
krishhna123: Added name and tax category in rest model too for tests

models/cart/src/custom-line-item/field-config.ts
krishhna123: added total price in fields also added adjusted name property according to the model

models/cart/src/custom-line-item/custom-line-item-draft/field-config.ts
krishhna123: I went through the docs again and found that money has same shape in both REST and GRAPHQL model for custom line item draft mode. It is different for custom line item model so I updated it there.

models/cart/src/custom-line-item/index.ts
krishhna123: Updated `index.js` file for `cart` package

models/cart/src/custom-line-item/builders.spec.ts
krishhna123: Added test for build function on compatibility builder

models/cart/src/custom-line-item/custom-line-item-draft/builders.spec.ts
krishhna123: Added test for build function on compatibility builder

krishhna123:
> Thanks for working on this migration, Krishna 🙇
> 
> I left some comments with some small things I believe we should take into consideration.
> 
> I also wanted to know your point of view regarding the guidelines we have in the docs folder. Do you think they're useful enough? Are you missing something on them?

Hello @CarlosCortizasCT , Thank you for the review and I have implemented your comments. Please have a look.

Also I went through `migrating-existing-model.md` document while working on it and it was helpful. However I took help of existing PRs as well to work on complex properties. 

What I struggled little bit with is shape of `Graphql` models if they are different from `REST`, e.g. name, Money. Also adding extra fields in `Graphql` model takes some work to go through types and understand what is missing. Maybe some guideline over that can be helpful.

CarlosCortizasCT:
> > Thanks for working on this migration, Krishna 🙇
> > I left some comments with some small things I believe we should take into consideration.
> > I also wanted to know your point of view regarding the guidelines we have in the docs folder. Do you think they're useful enough? Are you missing something on them?
> 
> Hello @CarlosCortizasCT , Thank you for the review and I have implemented your comments. Please have a look.
> 
> Also I went through `migrating-existing-model.md` document while working on it and it was helpful. However I took help of existing PRs as well to work on complex properties.
> 
> What I struggled little bit with is shape of `Graphql` models if they are different from `REST`, e.g. name, Money. Also adding extra fields in `Graphql` model takes some work to go through types and understand what is missing. Maybe some guideline over that can be helpful.

Thanks for the feedback, Krishna 🙇 

Can I ask you, as a follow-up, to open another PR and add to the guidelines docs whatever you think it's missing that can help others in the future? 🙏

models/cart/src/custom-line-item/field-config.ts
```diff
@@ -0,0 +1,66 @@
...
+} from '@commercetools-test-data/core';
+import { customLineItemPriceMode } from '../cart/constants';
+import { TCustomLineItemRest, TCustomLineItemGraphql } from './types';
+
+const commonFieldsConfig = {
+  id: fake((f) => f.string.uuid()),
+  key: fake((f) => f.string.uuid()),
+  taxedPrice: null,
+  taxedPricePortions: [],
+  totalPrice: fake(() => CentPrecisionMoney.random()),
```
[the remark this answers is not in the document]
krishhna123: Yes you are right! Moved property to respective models

models/cart/src/custom-line-item/custom-line-item-draft/field-config.ts
```diff
@@ -0,0 +1,40 @@
...
+export const restFieldsConfig: TModelFieldsConfig<TCustomLineItemDraftRest> = {
+  fields: {
+    ...commonFieldsConfig,
+  },
+};
+
+export const graphqlFieldsConfig: TModelFieldsConfig<TCustomLineItemDraftGraphql> =
+  {
+    fields: {
+      ...commonFieldsConfig,
```
YahiaElTai, line 38: for graphql, I think tax category needs to have taxCategoryRef as well and anything that could have reference expansion needs 2 fields for GraphQL one for reference expansion and one for reference, check out the Cart entity to see how the `postbuild` is done

YahiaElTai (approved):
Lovely

models/cart/src/custom-line-item/builders.spec.ts
```diff
@@ -0,0 +1,135 @@
...
+      }),
+    })
+  );
+};
+
+const validateRestFields = (model: TCustomLineItemRest) => {
+  validateCommonFields(model);
+  expect(model).toEqual(
+    expect.objectContaining({
+      name: expect.objectContaining({
```
krishhna123, line 89: Added step to validate shape of name field in rest model

models/cart/src/custom-line-item/builders.spec.ts
```diff
@@ -0,0 +1,135 @@
...
+        en: expect.any(String),
+        de: expect.any(String),
+        fr: expect.any(String),
+      }),
+      taxCategory: null,
```
krishhna123, line 94: validating if taxCategory is null (default value as its not mandatory field)

models/cart/src/custom-line-item/builders.spec.ts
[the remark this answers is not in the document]
krishhna123: I have added inline comments for these changes.

krishhna123:
> > > Thanks for working on this migration, Krishna 🙇
> > > I left some comments with some small things I believe we should take into consideration.
> > > I also wanted to know your point of view regarding the guidelines we have in the docs folder. Do you think they're useful enough? Are you missing something on them?
> > 
> > 
> > Hello @CarlosCortizasCT , Thank you for the review and I have implemented your comments. Please have a look.
> > Also I went through `migrating-existing-model.md` document while working on it and it was helpful. However I took help of existing PRs as well to work on complex properties.
> > What I struggled little bit with is shape of `Graphql` models if they are different from `REST`, e.g. name, Money. Also adding extra fields in `Graphql` model takes some work to go through types and understand what is missing. Maybe some guideline over that can be helpful.
> 
> Thanks for the feedback, Krishna 🙇
> 
> Can I ask you, as a follow-up, to open another PR and add to the guidelines docs whatever you think it's missing that can help others in the future? 🙏

Sure @CarlosCortizasCT ! I can update the guidelines docs in follow up PR

krishhna123:
> Nice one, @krishhna123 ! have any presets used in MC been added here, or are there plans to add any? If so, do we need a standardized format for them as well?

MC presets are mostly used by our unit and integration tests and they are very specific to those tests only so I think we are not planning to add here but whatever existing presets are available in test-data repo, we will migrate them to new format.

To handle presets in MC, I am working on migrating them to use new test-data model and moving them to application specific folders so that once model gets fully migrated to test-data repo, we can safely remove it from MC and still preserve presets. 

https://github.com/commercetools/merchant-center-frontend/pull/18548 is a PR that I created to migrate presets to application-orders folder

models/cart/src/custom-line-item/builders.spec.ts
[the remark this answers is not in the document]
CarlosCortizasCT: I'm sorry but I still don't understand.

We're populating the `nameAllLocales` and the `taxCategory` in the GraphQL model but we're not doing it in the REST one.

It does not seem the `nameAllLocales` is needed as it is already populated by default (in the `fields-config.ts` file).

For the `taxCategory`, if there's any reason for actually testing it, I don't get why we only test the GraphQL model for it.
It might be to make sure the reference is automatically populated as well which I understand, but then I would recommend to also populate and validate this property in the REST model for consistency (we validate the same common properties in both models).

models/cart/src/custom-line-item/builders.spec.ts
```diff
@@ -0,0 +1,135 @@
+import { Money, LocalizedString } from '@commercetools-test-data/commons';
```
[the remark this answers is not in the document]
krishhna123: Done

models/cart/src/custom-line-item/builders.spec.ts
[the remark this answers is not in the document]
krishhna123: I think I understand your point now.

I have removed populating `nameAllLocales` from tests implicitly because it is populated by default. 

Also I am populating `tax category` both in `REST` and `GRAPHQL` model now in the test as it has different shape in both. And as you mentioned correctly, I am populating `tax category` in `GRAPHQL` model to validate if references are populated properly or not.

PR #768, opened 2025-03-31
Product test data model migration

CarlosCortizasCT:
In this PR we migrate the `Product` and `ProductDraft` test data models to the latest architecture patterns following the guidelines provided [here](https://github.com/commercetools/test-data/blob/main/docs/guidelines/migrating-existing-model.md).
We've also created the `SelectionOfProduct` test data model as it's the one used in the `productSelectionRefs` property of the `Product` data model, which was missing.

There are too many changed files but that's due to the b2b and b2c presets.
Please do not pay attention to those files as they don't have relevant changes.

PR #773, opened 2025-04-08
Fix ProductDraft compatibility model export

CarlosCortizasCT:
This PR fixes an issue introduced in  #768 where we changed the way the `ProductDraft` model is exported.

The new way is not backwards compatible so we need to adjust what we've changed over there so consumers are not affected.

**NOTE**: Changes in the presets can be ignored. We just needed to adjust some import paths.

### DIFF
--- a/models/product/src/index.ts
+++ b/models/product/src/index.ts
@@ -1,7 +1,6 @@
 // Export types
 export * from './attribute/types';
 export * from './image/types';
-export * from './product/types';
 export * from './product-catalog-data/types';
 export * from './product-data/types';
 export * from './product-variant/types';
@@ -13,8 +12,8 @@ export * as AttributeDraft from './attribute/attribute-draft';
 export * as Image from './image';
 export * as ImageDraft from './image/image-draft';
 
-export * as Product from './product';
-export * as ProductDraft from './product/product-draft';
+export * from './product';
+export * from './product/product-draft';
 
 export * as ProductCatalogData from './product-catalog-data';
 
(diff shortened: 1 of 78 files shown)


models/product/src/index.ts
```diff
@@ -13,8 +12,8 @@ export * as AttributeDraft from './attribute/attribute-draft';
 export * as Image from './image';
 export * as ImageDraft from './image/image-draft';
 
-export * as Product from './product';
-export * as ProductDraft from './product/product-draft';
```
CarlosCortizasCT, line 17: We now let the models define what to export so it's more flexible.

models/product/src/product/index.ts
```diff
@@ -5,11 +5,10 @@ import {
   GraphqlModelBuilder,
   RestModelBuilder,
 } from './builders';
+import * as constants from './constants';
 import * as modelPresets from './presets';
 
 export * from './types';
-export * as constants from './constants';
```
CarlosCortizasCT, line 11: We're still namespacing the constants export but in a different way (see Product export below).

models/product/src/product/index.ts
```diff
@@ -5,11 +5,10 @@ import {
...
-export * from './product-draft';
```
CarlosCortizasCT, line 12: The `ProductDraft` is exported on its own.

models/product/src/product/index.ts
```diff
@@ -21,9 +20,11 @@ export const ProductGraphql = {
   presets: modelPresets.graphqlPresets,
 };
 
-export * as Product from '.';
```
CarlosCortizasCT, line 24: This way of exporting was quite misleading so I believe it's better to explicitly export an object so we can see what we are exporting.

PR #777, opened 2025-04-09
Migrating line items and line items draft models to new model

krishhna123:
This PR migrates `LineItems` and `LineItemsDraft` to new models.


models/cart/src/line-item/field-config.ts
```diff
@@ -0,0 +1,115 @@
...
+  addedAt: fake(addedAt),
+  custom: null,
+  lastModifiedAt: fake(lastModifiedAt),
+};
+
+export const restFieldsConfig: TModelFieldsConfig<TLineItemRest> = {
+  fields: {
+    ...commonFieldsConfig,
+    name: fake(() => LocalizedString.random()),
+    productSlug: fake((f) => LocalizedString.presets.ofSlugs()),
```
CarlosCortizasCT, line 53: I was wondering why do we populate the `productSlug` in the REST variant but not in the GraphQL one 🤔.

I believe we usually only populate the required fields here.

models/cart/src/line-item/field-config.ts
```diff
@@ -0,0 +1,115 @@
...
+export const graphqlFieldsConfig: TModelFieldsConfig<TLineItemGraphql> = {
+  fields: {
+    ...commonFieldsConfig,
+    name: null,
+    productSlug: null,
+    productType: fake(() => ReferenceGraphql.random().typeId('product-type')),
+    variant: fake(() => ProductVariant.ProductVariantGraphql.random()),
+    totalPrice: fake(() => Money.random()),
+    supplyChannel: fake(() => ChannelGraphql.random()),
+    distributionChannel: fake(() => ChannelGraphql.random()),
```
CarlosCortizasCT, line 71: Usually we only populate the required model properties in the base test-data model.

Consumers can always use custom presets if they need something different.

I will add this to the guidelines.

models/cart/src/line-item/line-item-draft/field-config.ts
```diff
@@ -0,0 +1,50 @@
...
+  inventoryMode: oneOf(...Object.values(inventoryMode)),
+  shippingDetails: null,
+  custom: null,
+};
+
+export const restFieldsConfig: TModelFieldsConfig<TLineItemDraftRest> = {
+  fields: {
+    ...commonFieldsConfig,
+    perMethodExternalTaxRate: [],
+    distributionChannel: fake(() => ReferenceRest.random().typeId('channel')),
```
CarlosCortizasCT, line 37: I don't think we should be populating optional props in this file..

CarlosCortizasCT (commented):
Thanks for working in this migration, Krishna 🙇 

I left some comments with some suggestions.

models/cart/src/line-item/field-config.ts
```diff
@@ -0,0 +1,115 @@
+import { ChannelGraphql } from '@commercetools-test-data/channel';
+import {
+  CentPrecisionMoney,
+  LocalizedString,
+  Money,
+  Price,
```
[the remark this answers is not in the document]
krishhna123: I checked again and it seems `Price` is not yet migrated to new models. I checked it under `commons`. Am I missing something 🤔

models/cart/src/line-item/field-config.ts
```diff
@@ -0,0 +1,115 @@
...
+export const restFieldsConfig: TModelFieldsConfig<TLineItemRest> = {
+  fields: {
+    ...commonFieldsConfig,
+    name: fake(() => LocalizedString.random()),
+    productSlug: fake((f) => LocalizedString.presets.ofSlugs()),
+    productType: fake(() => ReferenceRest.random().typeId('product-type')),
+    variant: fake(() => ProductVariant.ProductVariantRest.random()),
+    totalPrice: fake(() => CentPrecisionMoney.random()),
+    supplyChannel: fake(() => ReferenceRest.random().typeId('channel')),
+    distributionChannel: fake(() => ReferenceRest.random().typeId('channel')),
```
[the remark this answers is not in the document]
krishhna123: Removed optional fields and building them in tests to validate if they are properly populated.

models/cart/src/line-item/field-config.ts
krishhna123: Removed optional fields and building them in tests to validate if they are properly populated

models/cart/src/line-item/field-config.ts
krishhna123: Removed optional fields and building them in tests to validate if they are properly populated

models/cart/src/line-item/line-item-draft/field-config.ts
krishhna123: Removed optional fields and building them in tests to validate if they are properly populated

krishhna123:
> Thanks for working in this migration, Krishna 🙇
> 
> I left some comments with some suggestions.

Thank you @CarlosCortizasCT for your review. My approach while populating option fields was to be aligned with the existing `generators` that we are deleting. Because there are some models like `LineItemDraft` which have only few required fields but I think its best to populate some important fields at least.
I have updated the models now according to your suggestions.

models/cart/src/line-item/field-config.ts
```diff
@@ -0,0 +1,115 @@
...
+  custom: null,
+  lastModifiedAt: fake(lastModifiedAt),
+};
+
+export const restFieldsConfig: TModelFieldsConfig<TLineItemRest> = {
+  fields: {
+    ...commonFieldsConfig,
+    name: fake(() => LocalizedString.random()),
+    productSlug: fake((f) => LocalizedString.presets.ofSlugs()),
+    productType: fake(() => ReferenceRest.random().typeId('product-type')),
```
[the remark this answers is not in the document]
krishhna123: Done

rajrdk (approved):

CarlosCortizasCT:
> hink its best to populate some important fields at l

Hi @krishhna123 

First of all, if we already had the non-migrated models populating some specific fields, we have to keep populating them in the migrated version. If we don't do that, we would be introducing a breaking change in the package which is something we must avoid.

Regarding the rule to follow about which props to populate, my point of view is still that only required fields should be populated and, in case we see it would be beneficial to have some optional props populated, I believe presets are the way to go.

Anyway, I think this is something that might worth a RFC about it so we can have an open discussion and recorded agreement.
Are you interested in working on such RFC?

models/cart/src/line-item/field-config.ts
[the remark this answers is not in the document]
CarlosCortizasCT: My bad.
Sorry about that 😞

models/cart/src/line-item/field-config.ts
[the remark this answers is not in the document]
krishhna123: Oh.. sorry.. Updated fields with presets.

krishhna123:
> > hink its best to populate some important fields at l
> 
> Hi @krishhna123
> 
> First of all, if we already had the non-migrated models populating some specific fields, we have to keep populating them in the migrated version. If we don't do that, we would be introducing a breaking change in the package which is something we must avoid.
> 
> Regarding the rule to follow about which props to populate, my point of view is still that only required fields should be populated and, in case we see it would be beneficial to have some optional props populated, I believe presets are the way to go.
> 
> Anyway, I think this is something that might worth a RFC about it so we can have an open discussion and recorded agreement. Are you interested in working on such RFC?

Sure @CarlosCortizasCT , I can work on the RFC and we can discuss on this topic.

models/cart/src/line-item/field-config.ts
```diff
@@ -0,0 +1,118 @@
...
+import { TLineItemRest, TLineItemGraphql } from './types';
+
+const [addedAt, lastModifiedAt] = createRelatedDates();
+
+const commonFieldsConfig = {
+  id: fake((f) => f.string.uuid()),
+  key: fake((f) => f.string.uuid()),
+  productId: fake((f) => f.string.uuid()),
+  productKey: fake((f) => f.string.uuid()),
+  productSlug: null,
```
CarlosCortizasCT, line 30: Since you mentioned this attributed was populated in the previous version of the test data model, we must populated in the migrated version to avoid breaking changes.
Bear in mind REST and GraphQL versions have different value types.

models/cart/src/line-item/field-config.ts
```diff
@@ -0,0 +1,118 @@
...
+  quantity: fake((f) =>
+    f.number.int({
+      min: 1,
+    })
+  ),
+  discountedPricePerQuantity: [],
+  taxedPrice: null,
+  taxedPricePortions: [],
+  state: [],
+  taxRate: null,
```
CarlosCortizasCT, line 41: Since you mentioned this attributed was populated in the previous version of the test data model, we must populated in the migrated version to avoid breaking changes.
Bear in mind REST and GraphQL versions have different value types.

models/cart/src/line-item/field-config.ts
```diff
@@ -0,0 +1,118 @@
...
+};
+
+export const graphqlFieldsConfig: TModelFieldsConfig<TLineItemGraphql> = {
+  fields: {
+    ...commonFieldsConfig,
+    name: null,
+    productType: fake(() => ReferenceGraphql.presets.productTypeReference()),
+    variant: fake(() => ProductVariantGraphql.random()),
+    totalPrice: fake(() => Money.random()),
+    supplyChannel: fake(() => ReferenceGraphql.presets.channelReference()),
```
CarlosCortizasCT, line 71: In the GraphQL version of this test data model the `supplyChannel` prop is not a reference but an actual `Channel`.

models/cart/src/line-item/field-config.ts
```diff
@@ -0,0 +1,118 @@
...
+    distributionChannel: fake(() =>
```
CarlosCortizasCT, line 72: In the GraphQL version of this test data model the `distributionChannel` prop is not a reference but an actual `Channel`.

models/cart/src/line-item/field-config.ts
```diff
@@ -0,0 +1,118 @@
...
+    supplyChannel: fake(() => ReferenceRest.presets.channelReference()),
+    distributionChannel: fake(() => ReferenceRest.presets.channelReference()),
+  },
+};
+
+export const graphqlFieldsConfig: TModelFieldsConfig<TLineItemGraphql> = {
+  fields: {
+    ...commonFieldsConfig,
+    name: null,
+    productType: fake(() => ReferenceGraphql.presets.productTypeReference()),
```
CarlosCortizasCT, line 68: In the GraphQL version of this test data model the `productType` prop is not a reference.

models/cart/src/line-item/line-item-draft/field-config.ts
```diff
@@ -0,0 +1,46 @@
...
+  addedAt: fake(addedAt),
+  externalPrice: fake(() => MoneyDraft.random()),
+  externalTotalPrice: null,
+  externalTaxRate: null,
+  inventoryMode: oneOf(...Object.values(inventoryMode)),
+  shippingDetails: null,
+  distributionChannel: null,
+  supplyChannel: null,
+  custom: null,
+};
```
CarlosCortizasCT, line 33: In order to avoid potential breaking changes, we should use the same values patterns we were using in the non-migrated model and populate the same properties:
```js
{
    productId: fake((f) => f.string.uuid()),
    sku: fake((f) => `${f.lorem.word()}-${f.string.alphanumeric(3)}`),
    quantity: fake((f) =>
      f.number.int({
        min: 1,
      })
    ),
    variantId: sequence(),
    supplyChannel: fake(() => ReferenceDraft.random().typeId('channel')),
    distributionChannel: fake(() => ReferenceDraft.random().typeId('channel')),
    externalTaxRate: null,
    externalPrice: fake(() => MoneyDraft.random()),
    externalTotalPrice: null,
    custom: null,
    inventoryMode: oneOf(...Object.values(inventoryMode)),
    shippingDetails: null,
    addedAt: fake(addedAt),
  },
  ```

File: models/cart/src/line-item/field-config.ts (hunk 9 above)
Comment (krishhna123) [reply models/cart/src/line-item/field-config.ts]: Updated it to `Channel` model

File: models/cart/src/line-item/field-config.ts (hunk 10 above)
Comment (krishhna123) [reply models/cart/src/line-item/field-config.ts]: Updated it to `Channel` model

File: models/cart/src/line-item/field-config.ts (hunk 11 above)
Comment (krishhna123) [reply models/cart/src/line-item/field-config.ts]: Update it to `ProductType`

File: models/cart/src/line-item/field-config.ts (hunk 7 above)
Comment (krishhna123) [reply models/cart/src/line-item/field-config.ts]: Populated attribute values similar to previous version while making sure shape of REST and GraphQL is maintained.

File: models/cart/src/line-item/field-config.ts (hunk 8 above)
Comment (krishhna123) [reply models/cart/src/line-item/field-config.ts]: Populated attribute values similar to previous version while making sure shape of REST and GraphQL is maintained.

**Comment by CarlosCortizasCT** [comment]
@krishhna123 could you also please update the `Cart` readme file section for the `LineItem` option? ([reference](https://github.com/commercetools/test-data/tree/main/models/cart#lineitem)).

We should use examples using the new REST and GraphQL models only ([example](https://github.com/commercetools/test-data/tree/main/models/inventory-entry#usage)).

File: models/cart/README.md [hunk 13]
```diff
@@ -33,34 +33,38 @@ const emptyCartDraft = CartDraft.presets.empty().build<TCartDraft>();
 
 ```ts
 import {
```
krishhna123, line 35: Updated examples form CustomLineItem model too which is already migrated

models/cart/README.md
```diff
@@ -33,34 +33,38 @@ const emptyCartDraft = CartDraft.presets.empty().build<TCartDraft>();
...
 
 ## LineItem
 
 ```ts
 import {
-  LineItem,
-  LineItemDraft,
-  type TLineItem,
-  type TLineItemDraft,
+  LineItemRest,
```
krishhna123, line 53: Updated examples form LineItem model

krishhna123:
> @krishhna123 could you also please update the `Cart` readme file section for the `LineItem` option? ([reference](https://github.com/commercetools/test-data/tree/main/models/cart#lineitem)).
> 
> We should use examples using the new REST and GraphQL models only ([example](https://github.com/commercetools/test-data/tree/main/models/inventory-entry#usage)).

Updated readme with examples.

PR #779, opened 2025-04-10
Fix ProductVariant compatibility builder

CarlosCortizasCT:
We've found an issue in the `ProductVariant` compatibility builder related to the `attributesRaw` property when a GraphQL representation was requested as it was not populated.

I also took the change to refactor the way we export that model to be easier to reason about and maintain.

Most of the changes in the PR are consequence of the export refactor. Please take a look to the inline comments to find the relevant parts to review.

### DIFF
--- a/models/product/src/index.ts
+++ b/models/product/src/index.ts
@@ -3,7 +3,6 @@ export * from './attribute/types';
 export * from './image/types';
 export * from './product-catalog-data/types';
 export * from './product-data/types';
-export * from './product-variant/types';
 
 // Export models
 export * as Attribute from './attribute';
@@ -19,7 +18,7 @@ export * as ProductCatalogData from './product-catalog-data';
 
 export * as ProductData from './product-data';
 
-export * as ProductVariant from './product-variant';
+export * from './product-variant';
 export * from './product-variant/product-variant-draft';
 
 export * from './selection-of-product';
--- a/models/product/src/product-variant/builders.spec.ts
+++ b/models/product/src/product-variant/builders.spec.ts
@@ -63,7 +63,12 @@ const validateGraphqlModel = (
         }),
       ]),
       sku: expect.any(String),
-      attributesRaw: [],
+      attributesRaw: expect.arrayContaining([
+        expect.objectContaining({
+          name: expect.any(String),
+          __typename: 'RawProductAttribute',
+        }),
+      ]),
       __typename: 'ProductVariant',
     })
   );
--- a/models/product/src/product-variant/fields-config.ts
+++ b/models/product/src/product-variant/fields-config.ts
@@ -36,10 +36,10 @@ export const graphqlFieldsConfig: TModelFieldsConfig<TProductVariantGraphql> = {
       // @ts-expect-error The GraphQL mode does not have a `attributes` field but we need this logic for the compatibility builder
       model.attributesRaw = model.attributes.map((attribute) => ({
         ...attribute,
-        __typename: 'Attribute',
+        __typename: 'RawProductAttribute',
       }));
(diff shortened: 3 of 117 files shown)


models/product/src/product-variant/builders.spec.ts
```diff
@@ -63,7 +63,12 @@ const validateGraphqlModel = (
         }),
       ]),
       sku: expect.any(String),
-      attributesRaw: [],
```
CarlosCortizasCT, line 66: We now make sure we have at least one item in this list (as updated in the corresponding `fields-config` file).

models/product/src/product-variant/fields-config.ts
```diff
@@ -27,7 +27,20 @@ export const restFieldsConfig: TModelFieldsConfig<TProductVariantRest> = {
 export const graphqlFieldsConfig: TModelFieldsConfig<TProductVariantGraphql> = {
   fields: {
     ...commonFieldsConfig,
-    attributesRaw: [],
+    attributesRaw: fake(() => [Attribute.random()]),
```
CarlosCortizasCT, line 30: Since we are populating the REST representation with one item in the `attributes` list, it makes sense to also have one in the GraphQL side.

models/product/src/product-variant/index.ts
```diff
@@ -5,9 +5,6 @@ import {
 } from './builders';
 import * as modelPresets from './presets';
 
-export * from './product-variant-draft';
```
CarlosCortizasCT, line 8: The draft version is exported on its own.

models/product/src/product-variant/index.ts
```diff
@@ -23,5 +20,7 @@ export const ProductVariantGraphql = {
 /**
  * @deprecated Use `ProductVariantRest` or `ProductVariantGraphql` exported models instead of `ProductVariant`.
  */
-export const random = CompatProductVariantModelBuilder;
-export const presets = modelPresets.compatPresets;
+export const ProductVariant = {
```
CarlosCortizasCT, line 23: Cleaner way of exporting the compatibility builder.

models/product/src/index.ts
```diff
@@ -3,7 +3,6 @@ export * from './attribute/types';
 export * from './image/types';
 export * from './product-catalog-data/types';
 export * from './product-data/types';
-export * from './product-variant/types';
```
CarlosCortizasCT, line 6: These types are now exported from the `ProductVarian` export below.

models/product/src/product-variant/fields-config.ts
```diff
@@ -27,7 +27,20 @@ export const restFieldsConfig: TModelFieldsConfig<TProductVariantRest> = {
...
     __typename: 'ProductVariant',
   },
+  postBuild: (model, context) => {
+    // @ts-expect-error The GraphQL mode does not have a `attributes` field but we need this logic for the compatibility builder
+    if (context.isCompatMode && model.attributes) {
+      // @ts-expect-error The GraphQL mode does not have a `attributes` field but we need this logic for the compatibility builder
+      model.attributesRaw = model.attributes.map((attribute) => ({
+        ...attribute,
+        __typename: 'Attribute',
```
ahmehri, line 39: Should it be the type that was used in the `builders.spec.ts`?

```suggestion
        __typename: 'RawProductAttribute',
```

CarlosCortizasCT: You're totally right.
Thanks for spotting that, Ahmed 🙇 

I fixed it here: d447d968429b46280f45be5081140ffab19c4577

I also added tests to the presets so this kind of issue won't happen again.

ahmehri (approved):

kterry1 (approved):

PR #787, opened 2025-04-25
chore: update GraphQL schemas and their generated types

ragafus:
- Updates GraphQL schemas and their generated types
- Fixes missing GraphQL models props.

In Priceless we are going to start using a new entity, `DiscountGroup`, and we would like to add their models ina follow-up PR.


krishhna123 (approved):
Looks fine to me. Also it will enable us to add model for recurring orders

models/product-tailoring/src/product-tailoring-data/fields-config.ts
```diff
@@ -34,6 +34,7 @@ export const graphqlFieldsConfig: TModelFieldsConfig<TProductTailoringDataGraphq
       metaDescriptionAllLocales: null,
       metaKeywordsAllLocales: null,
       __typename: 'ProductTailoringData',
+      attributesRaw: null,
```
CarlosCortizasCT, line 37: This attribute is not defined as optional in the corresponding type so we should be populating it at least with an empty array.

PR #790, opened 2025-04-29
refactor(product): migrate remaining models and fix inconsistencies

emmenko:
Needed to unblock some work in https://commercetools.atlassian.net/browse/FEC-40

Migrate remaining models to new structure:
* `ProductCatalogData`
* `ProductData`
* `CategoryOrderHint` (new)
* `SearchKeyword` (new)
* `SearchKeywords` (new)

With that, some inconsistencies with fields in `ProductData` model have been fixed.


models/product/src/product-data/fields-config.ts
```diff
@@ -0,0 +1,124 @@
...
+    })),
+    description: fake(() => LocalizedString.random()),
+    slug: fake(() => LocalizedString.random()),
+    metaTitle: fake(() => LocalizedString.random()),
+    metaDescription: fake(() => LocalizedString.random()),
+    metaKeywords: fake(() => LocalizedString.random()),
+    masterVariant: fake(() => ProductVariantRest.random()),
+    variants: fake(() => [ProductVariantRest.random()]),
+    searchKeywords: fake(() => SearchKeywordsRest.random()),
+  },
```
emmenko, line 31: New rest model

models/product/src/product-data/fields-config.ts
```diff
@@ -0,0 +1,124 @@
...
+  postBuild: (model) => {
+    return {
+      categories: model.categories.map<CategoryReference>((category) => {
+        const expandedCategory = category as unknown as TCtpCategory;
+        return {
+          id: expandedCategory.id,
+          typeId: 'category',
+          obj: expandedCategory,
+        };
+      }),
```
emmenko, line 41: To ensure that the shape matches the "expanded reference" shape

models/product/src/product-data/fields-config.ts
```diff
@@ -0,0 +1,124 @@
...
+    name: null, // computed
+    nameAllLocales: fake(() => LocalizedString.random()),
+    searchKeyword: null, // computed
+    searchKeywords: fake(() => [SearchKeywordsGraphql.random()]),
+    skus: fake((f) => [`${f.lorem.word()}-${f.string.alphanumeric(3)}`]),
+    slug: null, // computed
+    slugAllLocales: fake(() => LocalizedString.random()),
+    variant: fake(() => ProductVariantGraphql.random()),
+    variants: fake(() => [ProductVariantGraphql.random()]),
+  },
```
emmenko, line 71: GraphQL representation. Notice that some fields are "computed" in the `postBuild` step.

models/product/src/product-data/types.ts
```diff
@@ -1,64 +1,10 @@
...
-// The base generator model. Consumers configure these fields.
-export type TProductData = Omit<ProductData, 'categories'> & {
-  categories: Array<Category>;
-  searchKeyword?: Array<SearchKeyword> | null;
-  allVariants: Array<ProductVariant>;
-  variant: ProductVariant;
-  skus: Array<String>;
-};
+export type TProductDataRest = ProductData;
+export type TProductDataGraphql = TCtpProductData;
```
emmenko, line 6: Relying 100% on the generated types

CarlosCortizasCT (commented):
We should also update the model README so the code snippets use the new models.

models/product/src/product-data/category-order-hint/fields-config.ts
```diff
@@ -0,0 +1,11 @@
+import { fake, type TModelFieldsConfig } from '@commercetools-test-data/core';
+import type { TCategoryOrderHintGraphql } from './types';
+
+export const graphqlFieldsConfig: TModelFieldsConfig<TCategoryOrderHintGraphql> =
+  {
+    fields: {
+      __typename: 'CategoryOrderHint',
+      categoryId: fake((f) => f.string.uuid()),
+      orderHint: fake(() => String(Math.random())),
```
ismaelocaramelo, line 9: As far as I saw, `orderHint` should be between 0 and 1 and what it has been using is this `fake((f) => f.number.float({ min: 0.01, max: 0.99 }).toString())`

ismaelocaramelo (commented):
Just added a small comment

models/product/src/product-data/category-order-hint/fields-config.ts
emmenko: `Math.random()` returns a value between 0 and 1

emmenko:
The PR is ready for review again. The changes have been addressed in the commit https://github.com/commercetools/test-data/pull/790/commits/dd2f85fa6f06b984a26b8d8213963be386e80393 and the following ones.

models/product/src/product-data/builders.spec.ts
```diff
@@ -0,0 +1,299 @@
...
+        }),
+      ]),
+      masterVariant: expect.objectContaining({ id: expect.any(Number) }),
+      variant: expect.objectContaining({ id: expect.any(Number) }),
+      variants: expect.arrayContaining([
+        expect.objectContaining({ id: expect.any(Number) }),
+      ]),
+      allVariants: expect.arrayContaining([
+        expect.objectContaining({ id: expect.any(Number) }),
+      ]),
```
CarlosCortizasCT, line 140: When validating GraphQL models, we usually rely on the `__typename` attribute. That should be enough from the model point of view.
We are using that attribute in other properties of this model so I was wondering why we don't use it for all of them.

models/product/src/product-data/presets/boring-generic-milk-product-data.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+        expect.objectContaining({
+          locale: 'en',
+          value: 'boring-generic-milk-slug',
+        }),
+      ]),
+    })
+  );
+}
+
+describe('Product "boring generic milk" presets', () => {
```
CarlosCortizasCT, line 55: It would be great to also have tests for the compatibility presets.

models/product/src/product-data/presets/happy-cow-milk-product-data.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+        expect.objectContaining({
+          locale: 'en',
+          value: 'happy-cow-milk-slug',
+        }),
+      ]),
+    })
+  );
+}
+
+describe('Product "happy cow milk" presets', () => {
```
CarlosCortizasCT, line 54: It would be great to also have tests for the compatibility presets.

models/product/src/product-data/presets/boring-generic-milk-product-data.spec.ts
emmenko: https://github.com/commercetools/test-data/pull/790/commits/36a06047a64fd86c14b636fd044b155a0205289f

Note: the test for graphql compat preset is marked as skipped. The main problem is that the the compat builder uses e.g. `name()` but the graphql builder expects `nameAllLocales()`. This leads to the name data to be randomly generated as the object passed to `name()` gets ignored. Therefore, the expected graphql object doesn't match.

We can discuss this case later if you want.

PR #792, opened 2025-04-30
fix: preset package publish config

emmenko:
Follow up of #697 

image

We need to explicitly mark the package as public. That should do the trick.


CarlosCortizasCT (approved):
Thanks 🙇

PR #799, opened 2025-05-07
Fix Product model exports

CarlosCortizasCT:
We have an issue with the way we export some of the models in the `@commercetools-test-data/product` package coming from a recent migration.

This kind of imports:
```
import { ProductCatalogData } from '@commercetools-test-data/product';
```
do not work as we're namespacing those models twice.

Consumers would need to do something like this:
```
import { ProductCatalogData } from '@commercetools-test-data/product';

const model = ProductCatalogData.ProductCatalogData.random().build();
```

In this PR we're fixing the exports and also updating the README to reflect the new GraphQL and REST models.

### DIFF
--- a/.changeset/bright-walls-drop.md
+++ b/.changeset/bright-walls-drop.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-test-data/product': patch
+---
+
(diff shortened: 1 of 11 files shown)


tdeekens (approved):

emmenko (approved):
Thanks

PR #801, opened 2025-05-07
FEC-236 - Bundle single package correctly

emmenko:
# Background

We are bundling each model as a separate package. However, the new single package with multiple entrypoints has to be bundled correctly with all the models.

There must be no dependency of each individual package `@commercetools-test-data/*`.

To be backwards compatible during the transition phase we want to still publish the individual packages in addition to the single package.
One challenge is to ensure that the single package is bundled correctly.

# Goals

The new single package `@commercetools/composable-commerce-test-data` must bundle all test data models correctly, without any dependencies of the previous individual packages.

# Proposal

We tried first to use path aliases for `@commercetools-test-data/*` to keep the source files in the their current places (e.g. `./models/*`) but refer to them from the single package.
However, Preconstruct rightfully complains that:
* Relative import paths should not point outside of the package itself ([ref](https://preconstruct.tools/errors#all-relative-imports-in-a-package-should-only-import-modules-inside-of-their-package-directory-but-path-is-importing-import)).
* Importing package names must have those packages listed in dependencies ([ref](https://preconstruct.tools/errors#import-is-imported-by-path-but-the-package-is-not-specified-in-dependencies-or-peerdependencies)).

This obviously is a dead end.

What we need essentially is:
* For the new package to have the source files within its `src` folder.
* For the new package to not have any dependencies of `@commercetools-test-data/*`.
* For the new package's source files that rely on other models to use relative paths.
* For the new package to have the correct list of entry points.

We can therefore have a script that:
* Copies all source files from `./models`, `./core`, `./utils`, `./graphql-types` into the new package's `src` folder.
* Changes all imports of `@commercetools-test-data/*` into relative imports.
* Verifies that the entrypoints are correctly defined.

(The script was generated with the help of Cursor).

The copied source files are git ignored, so they are only generated during the build step.

PS: I also opted to rename the package folder from `./presets/all-packages` to `./package`. Eventually this will be renamed to `src` once we remove the individual packages setup (FEC-185).


generators/src/new-test-model/new-test-model.generator.ts
```diff
@@ -1,6 +1,8 @@
 import { execSync } from 'node:child_process';
-import { existsSync, mkdirSync, writeFileSync } from 'node:fs';
+import { constants } from 'node:fs';
+import { access, mkdir, writeFile, readFile, cp } from 'node:fs/promises';
```
emmenko, line 3: Since we are adjusting this file, we might as well improve a bunch of other things (e.g. using async functions).

generators/src/new-test-model/new-test-model.generator.ts
```diff
@@ -1,6 +1,8 @@
...
 import { dirname, join } from 'node:path';
+import { findRoot } from '@manypkg/find-root';
```
emmenko, line 5: Easier way to determine the repository's root folder

generators/src/new-test-model/new-test-model.generator.ts
```diff
@@ -133,123 +133,135 @@ export const newTestModelGenerator: CodeGenerator = {
...
-        'all-packages'
+      const standalonePackagePackageJsonPath = join(
+        rootDir,
+        'package',
+        'package.json'
       );
-      // Generate the re-export proxy file
-      writeFileSync(
-        join(presetDirectoryPath, 'src', `${modelCodename}.ts`),
-        `export * from '@commercetools-test-data/${modelCodename}';`
```
emmenko, line 211: This has been removed given the changes with the package's structure (we're now copying the source code of the model itself).

generators/src/new-test-model/new-test-model.generator.ts
```diff
@@ -133,123 +133,135 @@ export const newTestModelGenerator: CodeGenerator = {
...
       );
       // Update package.json
-      const presetPackageJson = (
-        await import(join(presetDirectoryPath, 'package.json'))
-      ).default;
-      presetPackageJson.files.push(modelCodename);
-      presetPackageJson.preconstruct.entrypoints.push(`${modelCodename}.ts`);
-      presetPackageJson.dependencies[
-        `@commercetools-test-data/${modelCodename}`
-      ] = 'workspace:*';
```
emmenko, line 221: Also not needed anymore

generators/src/new-test-model/new-test-model.generator.ts
```diff
@@ -133,123 +133,135 @@ export const newTestModelGenerator: CodeGenerator = {
...
+        JSON.stringify(standalonePackageJson, null, 2),
+        { encoding: 'utf-8' }
+      );
+
+      console.log(`Copying generated model into the standalone package`);
+      // Copy the generated model into the standalone package
+      await cp(
+        join(outputPath, 'src'),
+        join(rootDir, 'package', 'src', modelCodename),
+        { recursive: true }
```
emmenko, line 234: This is new: we're copying the source code of the new model so that Preconstruct can compile the entrypoint code.

models/channel/src/channel-draft/builder.spec.ts
```diff
@@ -1,8 +1,5 @@
 import { CustomFieldBooleanType } from '@commercetools-test-data/type';
-import {
-  getDefaultGraphqlLocalesExpect,
-  getDefaultRestLocalesExpect,
-} from '@commercetools-test-data/utils/src/test-utils';
```
emmenko, line 5: Wrong import

package/package.json
```diff
@@ -0,0 +1,137 @@
...
+    "store",
+    "tax-category",
+    "type",
+    "user",
+    "utils",
+    "zone"
+  ],
+  "preconstruct": {
+    "entrypoints": [
+      "./associate-role/index.ts",
```
emmenko, line 76: Since we're copying the source code of each model, the entrypoints should point to the `index.ts` of each model's folder.

package/package.json
```diff
@@ -0,0 +1,137 @@
...
+    "@babel/runtime": "^7.17.9",
+    "@babel/runtime-corejs3": "^7.17.9",
+    "@commercetools-frontend/application-config": "^22.10.0",
+    "@commercetools-frontend/constants": "^22.10.0",
+    "@commercetools/platform-sdk": "8.8.0",
+    "@faker-js/faker": "^9.0.0",
+    "@types/lodash": "^4.17.0",
+    "lodash": "^4.17.21",
+    "omit-deep": "^0.3.0"
+  }
```
emmenko, line 136: These are all aggregated dependencies from all models.

scripts/copy-sources.mjs
emmenko, line 1: Nit: This file has been mostly generated via Cursor but it does what it needs to.

CarlosCortizasCT:
I think after checking out this change developers will need to remove some folders from their repository, mainly the built bundles from the previous directories.

Can we at least leave some instructions in the description of this PR so developers know what they need to after after they check out the new version of the repo?

.eslintrc.js
```diff
@@ -5,6 +5,7 @@ module.exports = {
   // TODO: use the new `eslint-config-node`?
   extends: ['@commercetools-frontend/eslint-config-mc-app'],
   rules: {
+    'import/export': 'off', // To avoid the error: `RangeError: Maximum call stack size exceeded`
```
CarlosCortizasCT, line 8: Where is the `RangeError: Maximum call stack size exceeded` error coming from?

CarlosCortizasCT (commented):
Can we have a changeset? 🙏

.eslintrc.js
emmenko: It's from CI when running linters / tests. The CI workflow on PRs is configured with `--changedSince=main`, guessing due to the amount of sample data files to process.

Once we decouple the sample data, we can re-adjust some of those things.

CarlosCortizasCT: Ah! If I understand correctly, it's because this PR has lots of files changed but it can be reverted afterwards, right?
Can we leave a TODO comment so we don't forget about it? 🙏

emmenko: Yep https://github.com/commercetools/test-data/pull/801/commits/ce54ae76758af750c9d379c3d49de6aab4750551

CarlosCortizasCT:
> I think after checking out this change developers will need to remove some folders from their repository, mainly the built bundles from the previous directories.
> 
> Can we at least leave some instructions in the description of this PR so developers know what they need to after after they check out the new version of the repo?

@emmenko can you please take a look at this?

CarlosCortizasCT (approved):
Thanks a lot for this work 🙇

tdeekens (approved):
My light bulks flickered when opening some diffs. Otherwise looks good.

leventekobor (approved):

PR #803, opened 2025-05-08
fix(product): rest representation of categoryOrderHints is an object, not an array

emmenko:
Leftover of #790


Rombelirk (approved):

CarlosCortizasCT (approved):
Thanks for the fix 👍

PR #807, opened 2025-05-08
FEC-183 - Add codemod script to migrate imports

emmenko:
To keep things as simple as possible, we don't really need to have a proper codemod with AST parsing.

At the end of the day, it's a simple find and replace.

We can copy the script as-is in the repository that we want to run it for, the execute it (optionally providing a specific folder path).

### DIFF
--- a/scripts/transform-imports.mjs
+++ b/scripts/transform-imports.mjs
@@ -28,6 +28,11 @@ import { readFileSync, writeFileSync } from 'node:fs';
 import { readdir } from 'node:fs/promises';
 import { join } from 'node:path';
 
+const PACKAGE_JSON = 'package.json';
+const TEST_DATA_SCOPE = '@commercetools-test-data/';
+const NEW_TEST_DATA_PKG = '@commercetools/composable-commerce-test-data';
+const NEW_TEST_DATA_VERSION = '11.0.0';
+
 const transformFile = (filePath) => {
   const content = readFileSync(filePath, 'utf8');
   const transformed = content.replace(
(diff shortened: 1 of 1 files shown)


scripts/transform-imports.mjs
```diff
@@ -0,0 +1,77 @@
...
+/**
+ * Transform Imports Script
+ *
+ * This script helps migrate imports from @commercetools-test-data/* to @commercetools/composable-commerce-test-data/*
+ *
+ * Usage:
+ *   ./scripts/transform-imports.mjs [path]
+ *
+ * Arguments:
+ *   path    Path to the directory containing files to transform (default: current directory)
```
CarlosCortizasCT, line 12: Does the script works recursively?

scripts/transform-imports.mjs
```diff
@@ -0,0 +1,77 @@
...
+      files.push(...(await findFiles(fullPath)));
+    } else if (entry.name.match(/\.(ts|tsx|js|jsx)$/)) {
+      files.push(fullPath);
+    }
+  }
+
+  return files;
+};
+
+const main = async () => {
```
CarlosCortizasCT, line 68: The codemod should also take into account updating the `package.json` file of the corresponding application since consumers will need to remove the previous individual packages and add new one.

emmenko: Good point, I'll look into that.

scripts/transform-imports.mjs
emmenko: Yes, see `findFiles` function.

scripts/transform-imports.mjs
emmenko: Now the `package.json` is also updated: https://github.com/commercetools/test-data/pull/807/commits/23a7a677ee29ce3dd14ac8bf00fd3d16320019ba

image

scripts/transform-imports.mjs
```diff
@@ -0,0 +1,127 @@
...
+ */
+
+import { readFileSync, writeFileSync } from 'node:fs';
+import { readdir } from 'node:fs/promises';
+import { join } from 'node:path';
+
+const PACKAGE_JSON = 'package.json';
+const TEST_DATA_SCOPE = '@commercetools-test-data/';
+const NEW_TEST_DATA_PKG = '@commercetools/composable-commerce-test-data';
+const NEW_TEST_DATA_VERSION = '11.0.0';
```
emmenko, line 34: We'll need to put the correct version here. I just assume we'll be releasing v11 as the next one.

scripts/transform-imports.mjs
CarlosCortizasCT: Can we include in the script a call to `pnpm install` if any `package.json` file has been updated so it does all the work for the consumers?

emmenko: I think we are overcomplicating this.

CarlosCortizasCT: Why do you think so?

I believe the goal of this script is make it as easy as possible for consumers to migrate to the new single package in their repository.

The more the script can do, the easier would be for consumers to go through this breaking change version and we increase trust from developers on this kind of big changes.

emmenko: The codemod does what it needs to: refactor / migrate the code. Installing the deps is something extra that it's not really a concern of a codemod.

As long as the migration instructions are clear, I don't see a need to attempt installing the deps.

Of course we can include that in the script, but then do we need to account for the package manager or keep it hardcoded to PNPM?
I don't think it helps much in this scenario.

PS: wouldn't it be simpler if the chapter runs the migration? Or are you expecting teams to do that?

CarlosCortizasCT: Well, I see that we don't align in how much help this script should provide.

From my point of view, if we don't include what you call an extra step, whoever runs the script needs to know they would need to that and I believe that's something that can be automated so the migration would be simpler to execute.

Do we need to account for the package manager? I think it would be great. We can just check which lock file the repo has and based the package manager selection on that (keeping it within the list of `pnpm`, `yarn` and`npm`.

> wouldn't it be simpler if the chapter runs the migration? Or are you expecting teams to do that?
Yes, that's something we can do although is not strictly necessary and perhaps this script can be used by external Custom Applications developers.

If you still think adding that part to the script is too much, may I ask at least you add detailed instructions on how to use the script in the changelog of the new major version of the single package?

emmenko: Here you go: https://github.com/commercetools/test-data/pull/807/commits/e88865a60582036a12a86a06f0e893d3d5446b0d

CarlosCortizasCT: Many thanks 🙇

CarlosCortizasCT (approved):
I gave it a try in the `audit-log` repository and it worked with no problems 👍

Rombelirk (approved):

leventekobor (approved):

PR #809, opened 2025-05-12
FEC-237: to migrate product selection to new model and add product of selection submodel

Rombelirk:
This PR migrates Product Selection to the new model,
also it adds Product Of Selection submodel to it.


standalone/src/models/product-selection/fields-config.ts
```diff
@@ -0,0 +1,54 @@
...
+      productRefs: null,
+    },
+    postBuild: (model) => {
+      const name = model.nameAllLocales
+        ? LocalizedString.resolveGraphqlDefaultLocaleValue(model.nameAllLocales)
+        : undefined;
+
+      //   const productRefs = ReferenceGraphql.presets
+      //     .productReference()
+      //     .buildGraphql();
```
Rombelirk, line 47: Not sure how to properly populate this field, can someone assist here?

standalone/src/models/store/product-selection-setting/product-selection-setting-draft/presets/sample-data-b2b/default-product-selection.ts
```diff
@@ -1,8 +1,6 @@
 import { KeyReferenceDraft } from '../../../../../commons';
-import {
-  ProductSelectionDraft,
-  type TProductSelectionDraft,
-} from '../../../../../product-selection';
+import { ProductSelectionDraft } from '../../../../../product-selection';
+import type { TProductSelectionDraft } from '../../../../../product-selection/types';
```
Rombelirk, line 3: I wasn't sure if I should also export the types from the index file to avoid this change

standalone/src/models/product-selection/product-selection-draft/presets/index.ts
```diff
@@ -1,6 +1,27 @@
...
 import sampleDataB2B from './sample-data-b2b';
 
-const presets = { empty, sampleDataB2B };
-
-export default presets;
+export const restPresets = {
+  empty: empty.restPreset,
+  sampleDataB2B: {
+    usMediumCustomersCatalog: sampleDataB2B.usMediumCustomersCatalog.restPreset,
+    defaultProductSelection: sampleDataB2B.defaultProductSelection.restPreset,
```
Rombelirk, line 8: I wasn't sure here how to go about this nested preset structure
`presets -> sampleDataB2B -> usMediumCustomersCatalog, defaultProductSelection`
is it fine to do it like this?

standalone/src/models/product-selection/product-selection-draft/presets/sample-data-b2b/default-product-selection.ts
```diff
@@ -1,9 +1,39 @@
 import { LocalizedStringDraft } from '../../../../commons';
-import type { TProductSelectionDraftBuilder } from '../../../types';
-import * as ProductSelectionDraft from '../../index';
+import type { TBuilder } from '../../../../../core';
+import { TCtpProductSelectionMode } from '../../../../../graphql-types';
```
Rombelirk, line 3: Should we now use relative paths instead of `@commercetools-test-data/graphql-types` here and for all the imports alike?

standalone/src/models/store/product-selection-setting/product-selection-setting-draft/presets/sample-data-b2b/default-product-selection.ts
CarlosCortizasCT: The product-selection index file should also export the types, yes.

standalone/src/models/product-selection/product-selection-draft/presets/sample-data-b2b/default-product-selection.ts
CarlosCortizasCT: At this moment we need to use relative paths but we will move to aliases once [this PR](https://github.com/commercetools/test-data/pull/814) is merged.

standalone/src/models/product-selection/fields-config.ts
CarlosCortizasCT: I would suggest to investigate the shape of that type in the GraphQL explorer to understand what's needed.
Perhaps we need some new sub-models.

Bear in mind that property is required and so we need to populate it with a value. In this case, the property seems to be an object with some results subprops.

Screenshot 2025-05-13 at 17 46 53
Screenshot 2025-05-13 at 17 47 23

standalone/src/models/product-selection/product-selection-draft/fields-config.ts
```diff
@@ -0,0 +1,25 @@
...
+    },
+  };
+
+export const graphqlFieldsConfig: TModelFieldsConfig<TProductSelectionDraftGraphql> =
+  {
+    fields: {
+      name: fake((f) => f.lorem.word()),
+      key: fake((f) => f.lorem.slug()),
+      custom: null,
+      mode: 'Individual',
```
CarlosCortizasCT, line 23: `name` property seem to have different share in the GraphQL type so we need to adjust here accordingly.

standalone/src/models/product-selection/product-selection-draft/presets/index.ts
CarlosCortizasCT: I pushed [this commit](https://github.com/commercetools/test-data/pull/809/commits/604ad75f87da3c09b69c863414ebd8c2d5dd4128) with a refactor for one of the presets. 
Perhaps you can take it as a basis for refactoring the other one.

standalone/src/models/product-selection/fields-config.ts
Rombelirk: So should I first create a submodel for `ProductOfSelection` and then proceed with populating this field here?
Or are we fine to leave it as `null` for now and then return and add this field as a follow-up once the sub-model is there? I mean, it's null in the current model so we will not break anything by leaving it so, right?
Screenshot 2025-05-14 at 13 27 17

CarlosCortizasCT: It is `null` in the current model but that's an issue because that means the test data model is generating an invalid object because the property is required.

I think we do need to create that submodel since we're already working on this test data model.

Rombelirk: The submodel is added ✅

.changeset/spicy-numbers-deny.md
```diff
@@ -0,0 +1,20 @@
...
+Added new submodel to the ProductSelection model:
+
+- `ProductOfSelection`
+
+You can use it like this:
+
+```ts
+import {
+  ProductOfSelectionGraphql,
+} from '@commercetools/composable-commerce-test-data/product-selection';
```
Rombelirk, line 14: Also, should it be exposed at all from the index file? Cause it's not supposed to be used outside of the parent model, or is it?

.changeset/spicy-numbers-deny.md
```diff
@@ -0,0 +1,20 @@
+---
+"@commercetools/composable-commerce-test-data": patch
```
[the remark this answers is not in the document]
CarlosCortizasCT: I still think this should be a patch versions for the reasons we've discussed in the last couple of dailies.

Also, I don't expect consumers to have tests expecting this being `null` as that's something I haven't seen. Test usually validate a value exists rather than the opposite (when working with data models).

.changeset/spicy-numbers-deny.md
CarlosCortizasCT: Yes they should be exposed. 
Remember consumers need the ability to populate the `productRefs` property and they should use the new models for that.

standalone/src/models/product-selection/product-of-selection/index.ts
```diff
@@ -0,0 +1,8 @@
+import { GraphqlModelBuilder } from './builders';
+import * as ProductOfSelectionPresets from './presets';
+export * from '../types';
+
+export const ProductOfSelectionGraphql = {
```
CarlosCortizasCT, line 5: Please leave a comment here explaining why we only have the GraphQL version of the model.

standalone/src/models/product-selection/product-of-selection/fields-config.ts
```diff
@@ -0,0 +1,18 @@
+import { type TModelFieldsConfig } from '@/core';
+import { ReferenceGraphql } from '@/models/commons';
+import type { TProductOfSelectionGraphql } from '../types';
+
+const commonFieldsConfig = {
+  productRef: ReferenceGraphql.presets.productReference().buildGraphql(),
+  product: null,
+  variantSelection: null,
+  variantExclusion: null,
+};
```
CarlosCortizasCT, line 10: No need to use a `common` constant as these props are only used once.
Please move them to the config object below.

standalone/src/models/product-selection/types.ts
```diff
@@ -3,24 +3,36 @@ import type {
...
-};
+
+// REST types
+export type TProductSelectionDraftRest = ProductSelectionDraft;
+export type TProductSelectionRest = ProductSelection;
+
+// GraphQL types
+export type TProductSelectionDraftGraphql = TCtpCreateProductSelectionDraft;
+export type TProductSelectionGraphql = TCtpProductSelection;
+export type TProductOfSelectionGraphql = TCtpProductOfSelection;
```
CarlosCortizasCT, line 28: The `ProductOfSelection` model should have its own `types.ts` file.

standalone/src/models/product-selection/builders.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+  expect(model).toMatchObject({
+    version: expect.any(Number),
+    id: expect.any(String),
+    createdAt: expect.any(String),
+    lastModifiedAt: expect.any(String),
+    lastModifiedBy: expect.objectContaining({
+      __typename: 'Initiator',
+      externalUserId: expect.any(String),
+      anonymousId: expect.any(String),
+      clientId: expect.any(String),
```
CarlosCortizasCT, line 58: There's no need to validate specific functions of nested models in a GraphQL one. Asserting the `__typename` property should be enough.

standalone/src/models/product-selection/types.ts
Rombelirk: I see. But why it's not the case for the draft model then?

standalone/src/models/product-selection/builders.spec.ts
Rombelirk: Makes sense, we already tested those fields in the submodel's test 👍🏻

standalone/src/models/product-selection/types.ts
CarlosCortizasCT: The draft model is a sibling of the "main" model. They relate to the very same entity in the APIs.
A submodel is a different model that is related to its parent.

standalone/src/models/product-selection/builders.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+  it('builds a GraphQL model', () => {
+    const graphqlModel = ProductSelectionGraphql.random().build();
+
+    validateGraphqlModel(graphqlModel);
+  });
+});
+
+describe('ProductSelection model compatibility builders', () => {
+  it('builds a default (REST) model', () => {
+    const restModel = ProductSelection.random().build();
```
rajrdk, line 98: Should be using `ProductSelectionRest` for building the rest model, right?

standalone/src/models/product-selection/builders.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+
+    validateRestModel(restModel);
+  });
+
+  it('builds a REST model', () => {
+    const restModel = ProductSelection.random().buildRest();
```
rajrdk, line 104: Same here, use `ProductSelectionRest` instead of `ProductSelection`.

standalone/src/models/product-selection/builders.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+
+    validateRestModel(restModel);
+  });
+
+  it('builds a GraphQL model', () => {
+    const graphqlModel =
+      ProductSelection.random().buildGraphql<TProductSelectionGraphql>();
```
rajrdk, line 111: Same here, use `ProductSelectionGraphql` instead of `ProductSelection`.

standalone/src/models/product-selection/builders.spec.ts
```diff
@@ -0,0 +1,115 @@
+import { TProductSelectionGraphql, TProductSelectionRest } from './types';
+import {
+  ProductSelection,
```
rajrdk, line 3: No need to import `ProductSelection`.

standalone/src/models/product-selection/builders.spec.ts
CarlosCortizasCT: Not in this case. 
In the tests we first validates the REST and GraphQL models but then we need to also validate the compatibility one, which is what's done here.

standalone/src/models/product-selection/builders.spec.ts
CarlosCortizasCT: It is actually needed.
Please see my comment below.

.changeset/spicy-numbers-deny.md
```diff
@@ -0,0 +1,20 @@
...
+---
+
+Added new submodel to the ProductSelection model:
```
CarlosCortizasCT, line 5:
```suggestion
Added new submodel to the ProductOfSelection model:
```

CarlosCortizasCT, line 5: We can also mentioned we migrated the `ProductSelection` model.

standalone/src/models/product-selection/product-of-selection/index.ts
```diff
@@ -0,0 +1,9 @@
+import { GraphqlModelBuilder } from './builders';
+import * as ProductOfSelectionPresets from './presets';
+export * from '../types';
+
+// We only export the GraphQL model for product-of-selection, as the REST model is not used in the current implementation.
```
CarlosCortizasCT, line 5:
```suggestion
// This model only exists in the GrahpQL API and that's why there's no REST version of it.
```

standalone/src/models/product-selection/product-of-selection/index.ts
```diff
@@ -0,0 +1,9 @@
+import { GraphqlModelBuilder } from './builders';
+import * as ProductOfSelectionPresets from './presets';
+export * from '../types';
```
CarlosCortizasCT, line 3: We must export the types of this model:
```suggestion
export * from './types';
```

standalone/src/models/product-selection/product-of-selection/builders.spec.ts
```diff
@@ -0,0 +1,21 @@
...
+describe('ProductOfSelection Builder', () => {
+  it('should build properties for the GraphQL representation', () => {
+    const graphqlModel = ProductOfSelectionGraphql.random().build();
+
+    expect(graphqlModel).toEqual(
+      expect.objectContaining({
+        productRef: expect.objectContaining({
+          __typename: 'Reference',
+          typeId: 'product',
+          id: expect.any(String),
```
CarlosCortizasCT, line 12: No need to check these properties. The `__typename` one is already enough.

standalone/src/models/product-selection/product-selection-draft/presets/empty.spec.ts
```diff
@@ -1,11 +1,44 @@
-import { TProductSelectionDraft } from '../../types';
-import empty from './empty';
+import type {
+  TProductSelectionDraftGraphql,
+  TProductSelectionDraftRest,
+  TProductSelectionDraft,
```
rajrdk, line 4: Unused `TProductSelectionDraft` import.

.changeset/spicy-numbers-deny.md
Rombelirk: But the model we are adding submodel to is called `ProductSelection`, `ProductOfSelection` is the name of the submodel.

CarlosCortizasCT: You're totally right. Sorry.

standalone/src/models/product-selection/product-of-selection/builders.spec.ts
Rombelirk: But why is that? I understand we don't need to test those when we are populating a field with a submodel cause we assume that those fields are already tested in the submodel and we just want to make sure the `__typename` is correct, but this is the test related to the submodel itself, and if we remove those fields how can we be confident that those fields even exist in this submodel?

Rombelirk: Ahhh, you're right! This is the fields of another submodel, thanks for noticing.

standalone/src/models/product-selection/product-selection-draft/presets/empty.spec.ts
Rombelirk: thanks for noticing 👍🏻

standalone/src/models/product-selection/builders.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+    }),
+    createdBy: expect.objectContaining({
+      __typename: 'Initiator',
+    }),
+    nameAllLocales: expect.arrayContaining([
+      expect.objectContaining({
+        __typename: 'LocalizedString',
+      }),
+    ]),
+    name: expect.any(String),
```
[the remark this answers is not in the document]
Rombelirk: Yes, we do.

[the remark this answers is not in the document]
CarlosCortizasCT: Generally speaking I would say yes, but it's just a matter of checking the GraphQL type definition of the GraphQL model we're working with:

image

standalone/src/models/product-selection/builders.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+      clientId: expect.any(String),
+      customer: expect.objectContaining({
+        id: expect.any(String),
+        typeId: 'customer',
+        obj: expect.objectContaining({
+          id: expect.any(String),
+        }),
+      }),
+    }),
+    createdBy: expect.objectContaining({
```
leventekobor, line 26: again just a question what, why dont we check for `attributedTo`?

CarlosCortizasCT: Usually in the REST models we don't test for all the properties of a nested model.
We just test one or two to make sure the shape looks like what we are expecting.

You need to understand the purpose here is just to make sure the property has the expected nested model. With GraphQL it's easy because we can rely on the `__typename` property. For REST, we use some kind of [duck typing](https://en.wikipedia.org/wiki/Duck_typing) for that. If a couple of properties match, that's enough. We think asserting for all of them is too much of a boilerplate and we rely on the nested models own tests to validate its structure.

leventekobor: thanks for the explanation @CarlosCortizasCT in both cases 🙇

standalone/src/models/product-selection/product-of-selection/index.ts
Rombelirk: Just to clarify, in this case we don't need to export types from the main index file (product-selection/index), only from this file, for the same reason we have `types` file separately for this submodel, right @CarlosCortizasCT ?

standalone/src/models/product-selection/product-of-selection/fields-config.ts
```diff
@@ -0,0 +1,14 @@
+import { type TModelFieldsConfig } from '@/core';
+import { ReferenceGraphql } from '@/models/commons';
+import type { TProductOfSelectionGraphql } from './types';
```
leventekobor, line 3: very nit but the `type` placement in the imports are inconsistent

Rombelirk: Sorry, what do you mean?

Rombelirk: Ah, I see. I think these particular imports were generated by CLI, so we might want to fix it there.

CarlosCortizasCT: The CLI does not know if the implementation of the file will require to import anything else from the `core` package so that's why it uses the safe way of importing a type.
Actually. the `type` keyword is optional and only used for some optimizations the bundler can do which I'm not sure they even apply in our case.

Please, focus on the relevant parts of the code 🙏

leventekobor: sorry still trying to figure out the relevant parts 🫡

standalone/src/models/product-selection/product-selection-draft/builders.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+    expect.objectContaining({
+      key: expect.any(String),
+      name: expect.any(Object),
+      mode: expect.any(String),
+      custom: expect.any(Object),
+    })
+  );
+};
+
+describe('ProductSelection model builders', () => {
```
leventekobor, line 24: also very nit and feel free to ignore if irrelevant but the model is `ProductSelectionDraft`

Rombelirk: that's a good one, thanks!

.changeset/spicy-numbers-deny.md
Rombelirk: Done ✅ 
Not sure about the wording being good enough though, please adjust as you see fit.

standalone/src/models/product-selection/product-selection-draft/fields-config.ts
```diff
@@ -0,0 +1,26 @@
...
+  TProductSelectionDraftRest,
+} from '../types';
+
+export const restFieldsConfig: TModelFieldsConfig<TProductSelectionDraftRest> =
+  {
+    fields: {
+      name: fake(() => LocalizedString.random()),
+      key: fake((f) => f.lorem.slug()),
+      custom: null,
+      mode: 'Individual',
```
leventekobor, line 14: since `standalone/src/models/product-selection/constants.ts` exists should we use those constants values here and in line 24?

Rombelirk: Yes, I think we should, thanks for pointing out.

leventekobor (approved):

CarlosCortizasCT (approved):
Thanks for working on this, Roman 🙇

PR #810, opened 2025-05-12
refactor(cart): group cart-discount into cart

emmenko:
https://commercetools.atlassian.net/browse/FEC-187

Group cart related models into `./cart`:

```
/cart
  /cart
  /cart-discounts
```


emmenko:
@CarlosCortizasCT Are there any other models that need to be moved into `./cart`?

CarlosCortizasCT:
> @CarlosCortizasCT Are there any other models that need to be moved into `./cart`?

You can find the whole new structure [in the RFC](https://commercetools.atlassian.net/wiki/spaces/FEC/pages/1232470022/RFC+Refactor+test-data+repository+published+packages).

I don't think we planned for any other model to go to this folder.

emmenko:
Yes I'm following the RFC. The `./cart` folder already contains lots of other models (which weren't listed in the RFC). That's why I'm asking 😊

CarlosCortizasCT:
> Yes I'm following the RFC. The `./cart` folder already contains lots of other models (which weren't listed in the RFC). That's why I'm asking 😊

If you take a look at the RFC you can see how the current `cart` folder was supposed to be a subfolder of the new parent `cart` folder which would also contain the `cart-discount` folder.
image

Let's agree on what to do here and the remaining work in our daily tomorrow.

babel.config.js
```diff
@@ -12,6 +12,9 @@ module.exports = {
           '@/core': './standalone/src/core',
           '@/core/test-utils': './standalone/src/core/test-utils',
           '@/graphql-types': './standalone/src/graphql-types',
+          '@/models/cart': './standalone/src/models/cart/cart',
+          '@/models/cart-discount':
+            './standalone/src/models/cart/cart-discount',
```
emmenko, line 17: @CarlosCortizasCT FYI: we need to adjust the path aliases as the folder hierarchy changed.

There are different options we can go fo:
1. Explicitly mapping nested models. We would need to adjust these mappings for all nested / grouped models.
2. For the grouped models, define a new proxy entry point (e.g. `./models/cart.ts` and `./models/cart-discounts.ts`) so that we retain the generic mapping of path aliases `"@/models/*": ["./standalone/src/models/*"]`
3. Change the mapping to `"@/models/*": ["./standalone/src/*"]` to rely on the package entry points. However, an import like `@/models/graphql-types` or `@/models/core` is then also possible which makes it weird.

CarlosCortizasCT: I vote against adding new aliases as it adds an extra layer of configuration I think we don't need.

If we regroup directories we just need to update the imports once and it seems that's something that can easily be done with "Find & Replace" IDE's feature.

The PR will be big anyway because we're moving around many files.

emmenko: Thanks, I would also prefer to avoid adding new aliases.

Then, is this way of importing ok with you?

```ts
import { Cart } from '@/models/cart/cart';
import { CartDiscount } from '@/models/cart/cart-discount';
```

CarlosCortizasCT: Internally in this repo, yes it's fine by me.

emmenko: https://github.com/commercetools/test-data/pull/810/commits/05f0f4150fd142986e3fc073abba47e8db243859

CarlosCortizasCT:
@emmenko would you be able to take a look at the typescript validation errors?

emmenko:
I looked into it but I don't know what the issue is to be honest. It only happens with the production build.

I'll continue to investigate later.

CarlosCortizasCT (approved):
Thanks a lot 👍

rajrdk (commented):
@CarlosCortizasCT, should we replace paths with alias? I noticed they're used in many files.

CarlosCortizasCT:
> @CarlosCortizasCT, should we replace paths with alias? I noticed they're used in many files.

That would be great but it's not strictly necessary.
Most of them are in sample data presets which will be moved out of this repository anyway ([reference](https://commercetools.atlassian.net/wiki/spaces/FEC/pages/1918861739/RFC+Decouple+sample+data+presets+from+test-data+repository)) so I wouldn't spend the time now.

tdeekens (approved):

rajrdk (approved):

PR #819, opened 2025-05-16
Adjust changeset version

CarlosCortizasCT:
We've incorrectly used a `major` version in a committed changeset so we're adjusting it in this PR.


tdeekens (approved):
I think this happens when people use the CLI locally. Cause then it steps through from major to fix version and at times the first suggestion (major) is selected.

PR #846, opened 2025-06-04
Ma/adjust product catalog data model

mario-priceless:
We're adjusting the Product data model so it does not populate the categories property by default since that's an optional one and those should not exist in a default model.

https://github.com/commercetools/merchant-center-frontend/blob/main/packages-shared/test-data/src/product-data/generator.js#L18

### DIFF
--- a/.changeset/pink-nights-invent.md
+++ b/.changeset/pink-nights-invent.md
@@ -3,4 +3,4 @@
 '@commercetools-test-data/generators': minor
 ---
 
-Updated the `Product data` and `ProductDraft` test models so it does not populate the `categories` property by default. That's an optional property and so the default version of the model should not populate it.
+Updated the `ProductData` and `ProductDraft` test models so it does not populate the `categories` property by default. That's an optional property and so the default version of the model should not populate it.
--- a/.changeset/pink-nights-invent.md
+++ b/.changeset/pink-nights-invent.md
@@ -3,4 +3,4 @@
 '@commercetools-test-data/generators': minor
 ---
 
-Updated the `Product data` test model so it does not populate the `categories` property by default. That's an optional property and so the default version of the model should not populate it.
+Updated the `Product data` and `ProductDraft` test models so it does not populate the `categories` property by default. That's an optional property and so the default version of the model should not populate it.
--- a/.changeset/pink-nights-invent.md
+++ b/.changeset/pink-nights-invent.md
@@ -1,0 +1,6 @@
+---
+'@commercetools/composable-commerce-test-data': minor
+'@commercetools-test-data/generators': minor
+---
+
+Updated the `Product data` test model so it does not populate the `categories` property by default. That's an optional property and so the default version of the model should not populate it.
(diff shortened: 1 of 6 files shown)


standalone/src/models/product/product/product-data/fields-config.ts
```diff
@@ -108,13 +103,6 @@ export const graphqlFieldsConfig: TModelFieldsConfig<TProductDataGraphql> = {
     const metaTitle = metaTitleAllLocales
       ? LocalizedString.resolveGraphqlDefaultLocaleValue(metaTitleAllLocales)
       : undefined;
-    const categoriesRef = model.categories.map<TReferenceGraphql<'category'>>(
-      (category) => ({
-        id: category.id,
-        typeId: 'category',
-        __typename: 'Reference',
-      })
-    );
```
CarlosCortizasCT, line 117: We should not remove this.

In case the model has been populated with some categories by the consumers, we still want to create the references automatically.

standalone/src/models/product/product/product/product-draft/fields-config.ts
```diff
@@ -17,7 +17,7 @@ const commonFieldsConfig = {
   slug: fake(() => LocalizedStringDraft.presets.ofSlugs()),
   key: fake((f) => f.lorem.slug()),
   description: fake(() => LocalizedStringDraft.random()),
-  categories: fake(() => [KeyReferenceDraft.presets.category()]),
+  categories: null,
```
mario-priceless, line 20: our on the model of the MC  [chore(productCatalogData): remove non-required initialization property](https://github.com/commercetools/test-data/pull/846/commits/b3e4610e6d6dd7e1f10dae8ad4c9c9621756427b)

.changeset/pink-nights-invent.md
```diff
@@ -0,0 +1,6 @@
+---
+'@commercetools/composable-commerce-test-data': minor
+'@commercetools-test-data/generators': minor
+---
+
+Updated the `Product data` test model so it does not populate the `categories` property by default. That's an optional property and so the default version of the model should not populate it.
```
CarlosCortizasCT, line 6: We're also changing the `ProductDraft` model so we have to note it here.

mario-priceless: Addressed ->  3f248b87709ccb172874ac11b37b9172f9d18473

.changeset/pink-nights-invent.md
```diff
@@ -0,0 +1,6 @@
+---
+'@commercetools/composable-commerce-test-data': minor
+'@commercetools-test-data/generators': minor
+---
+
+Updated the `Product data` and `ProductDraft` test models so it does not populate the `categories` property by default. That's an optional property and so the default version of the model should not populate it.
```
CarlosCortizasCT, line 6: Please, let's be consistent with the naming:
```suggestion
Updated the `ProductData` and `ProductDraft` test models so it does not populate the `categories` property by default. That's an optional property and so the default version of the model should not populate it.
```

CarlosCortizasCT (approved):
Thanks 👍

PR #849, opened 2025-06-06
Ma export types from product

mario-priceless:
we forgot to export the types from the product model


mario-priceless:
closed on priority of this https://github.com/commercetools/test-data/pull/848#pullrequestreview-2904340802

PR #852, opened 2025-06-06
CKT-2924 Recurrence policy models

krishhna123:


CarlosCortizasCT:
I'd like to ask some questions?
* There are only GraphQL builders for the new models. Are the entities only available in GraphQL APIs? Will they exist in REST?
* Have you used the CLI to generate the new models? I'm asking because I'm missing some changes in the `package.json` so if you used it, something needs to be fixed over there.

The way the models are exported could be improved to follow the pattern we're taking while migrating legacy models.
If that's ok for you, I can push a commit with the relevant changes which would be easier than explain the changes in the different files.

krishhna123:
> I'd like to ask some questions?
> 
> * There are only GraphQL builders for the new models. Are the entities only available in GraphQL APIs? Will they exist in REST?

Yes, as this feature is in early access, only GraphQL types are available and once feature is available in public beta, sdks will be updated and we can use rest types too.
> * Have you used the CLI to generate the new models? I'm asking because I'm missing some changes in the `package.json` so if you used it, something needs to be fixed over there.

Yes, I used CLI to generate models. 
> The way the models are exported could be improved to follow the pattern we're taking while migrating legacy models. If that's ok for you, I can push a commit with the relevant changes which would be easier than explain the changes in the different files.

Yes, you can push changes.

standalone/src/models/recurring-order/recurrence-policy/index.ts
```diff
@@ -0,0 +1,7 @@
+import { GraphqlModelBuilder } from './builders';
+import * as RecurrencePolicyPresets from './presets';
+
+export const RecurrencePolicyGraphql = {
```
CarlosCortizasCT, line 4: I would suggest to leave a comment explaining why there's no a REST model yet.

krishhna123:
> We're missing the changeset.
> 
Done
> Also, I would recommend [testing this change locally](https://github.com/commercetools/test-data?tab=readme-ov-file#steps-to-link-a-local-package) to make sure the setup is fine.

I have already tested this locally and set up is working fine.

.changeset/forty-roses-ask.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools/composable-commerce-test-data': patch
+---
+
+Added models for recurrencey policy and standard schedule entities
```
CarlosCortizasCT, line 5: Could we perhaps improve the changeset a bit so it's easier for changelog reader to know what's been introduced?

You can take a look at some of the current changesets in the repo ([example](https://github.com/commercetools/test-data/blob/main/.changeset/forty-carrots-check.md)).

krishhna123: Added new changeset. Thank you for suggestion.

CarlosCortizasCT (approved):
Thanks! 👍

PR #853, opened 2025-06-09
Update CustomApplication related models

CarlosCortizasCT:
When using the `CustomApplication` models in a consumer application I noticed the constants that we were exporting were not the right ones from that specific models and other two models should be exporting their constants.

Also, the `CustomApplication` default model was populating more fields than what we had before (in the consumer apps) so I think it's better to keep that one slim and I added a new preset which populates all the fields for better flexibility.


NickDevG (commented):
I added some comments about the list of scopes. The [whole list](https://github.com/commercetools/commercetools-docs/blob/main/websites/api/src/content/scopes.mdx) is updated not too often but for example soon scopes for recurring orders and payment methods will be added. 

Not sure if all scopes would make sense for the custom applications though.

CarlosCortizasCT:
> I added some comments about the list of scopes. The [whole list](https://github.com/commercetools/commercetools-docs/blob/main/websites/api/src/content/scopes.mdx) is updated not too often but for example soon scopes for recurring orders and payment methods will be added.
> 
> Not sure if all scopes would make sense for the custom applications though.

Thanks for the suggestion @NickDevG.

I'm not sure this list should be always up-to-date as they seem to be just for the sake of having some values in the model's props.

Since we're just migrating the models, I'm using the same values that was already there and they seem to be enough.

PR #857, opened 2025-06-10
Create PimSearchListView test data model

CarlosCortizasCT:
## Description

We're introducing a new model named `PimSearchListView` which is the first one associated with the "My views" Merchant Center functionality so we've created a new package entry point (`@commercetools/composable-commerce-test-data/my-view`) we can use to aggregate all the other use cases we will need in the future.

This is how the new model could be used:
```ts
import { PimSearchListViewGraphql } from '@commercetools/composable-commerce-test-data/my-view';

const model = PimSearchListViewGraphql.random().build();
```

### DIFF
--- a/standalone/src/models/my-view/my-view-sort/builders.spec.ts
+++ b/standalone/src/models/my-view/my-view-sort/builders.spec.ts
@@ -8,7 +8,7 @@ describe('MyViewSort Builder', () => {
     expect(graphqlModel).toEqual(
       expect.objectContaining({
         key: expect.any(String),
-        order: SortOrder.ASCENDING,
+        order: SortOrder.Asc,
         __typename: 'Sort',
       })
     );


NickDevG (approved):
I was thinking about exporting each specific view as an object, but since we only have GraphQL it makes no sense. 

Looks good!

standalone/src/models/my-view/pim-search-list-view/builders.spec.ts
```diff
@@ -0,0 +1,29 @@
+import { PimSearchListViewGraphql } from './index';
+
+describe('PimSearchListView Builder', () => {
+  it('should build properties for the GraphQL representation', () => {
+    const graphqlModel = PimSearchListViewGraphql.random().build();
```
krishhna123, line 5: nit: Can you also build `sort` and `table` for this view so that we can validate if those views are also built properly when called from other view?

PR #862, opened 2025-06-12
Update postBuild callback behaviour

CarlosCortizasCT:
## Description

`postBuild` is a callback function that can be passed to a `TModelFieldsConfig` configuration object so we can perform some required transformations after a model has been built.
Most of the time we use this callback to generate computed values. Example:
```ts
export const graphqlFieldsConfig: TModelFieldsConfig<TCustomLineItemGraphql> = {
  fields: { ... },
  postBuild: (model) => {
    const name = model.nameAllLocales
      ? LocalizedString.resolveGraphqlDefaultLocaleValue(model.nameAllLocales)
      : model.nameAllLocales;
    const taxCategoryRef = model.taxCategory
      ? ReferenceGraphql.presets
          .taxCategoryReference()
          .id(model.taxCategory.id)
          .buildGraphql()
      : model.taxCategory;

    return {
      name,
      taxCategoryRef,
    };
  },
};
```

The object returned from this callback was merged into the already built object the builder had generated.

This was working well so far but we've found a use case where we need to replace a property from the generated model. It is required for a compatibility builder since the REST model has a property named `attributes` but the GraphQL one is named `attributeDefinitions`.

For such cases, we need to be able to manipulate the generated object so we're changing what we expect from the `postBuild` callback.
Instead of expecting a partial model object which will be merged to the previously generated one, now the function expects the full generated object with the changes applied.

The change is not super impactful as the function already receives the full generated object as its first parameter.
This is how we would update the previous example:
```ts
export const graphqlFieldsConfig: TModelFieldsConfig<TCustomLineItemGraphql> = {
  fields: { ... },
  postBuild: (model) => {
    const name = model.nameAllLocales
      ? LocalizedString.resolveGraphqlDefaultLocaleValue(model.nameAllLocales)
      : model.nameAllLocales;
    const taxCategoryRef = model.taxCategory
      ? ReferenceGraphql.presets
          .taxCategoryReference()
          .id(model.taxCategory.id)
          .buildGraphql()
      : model.taxCategory;

    return {
      ...model, // <---- WE JUST NEED TO ADD THE ALREADY GENERATED PROPS
      name,
      taxCategoryRef,
    };
  },
};
```

NOTE FOR REVIEWERS:
- The first commit updates the `core` module to change the `postBuild` behaviour
- The second commits updates all the current `postBuild` usages
- The third commit updates the docs

### DIFF
--- a/.changeset/public-experts-drive.md
+++ b/.changeset/public-experts-drive.md
@@ -1,0 +1,8 @@
+---
+'@commercetools/composable-commerce-test-data': minor
+---
+
+We're changing how the internal `postBuild` callback used in the models definitions work.
+You can read more about this function in [the docs](https://github.com/commercetools/test-data/blob/main/docs/contributing/test-data-models-overview.md#fields-config).
+
+This refactor is internal and does not have any impact in the consumers.


CarlosCortizasCT:
> Looks good to me, now we have to check for any `postBuild` in open PRs

Thanks for pointing that out.

We currently have two open PRs (one is in `draft` yet) which would be affected.
I'll leave a comment over there but if we merge this PR before those, they will fail in the typecheck step so they won't be merged until they are updated.

YahiaElTai (approved):
looks good!

PR #864, opened 2025-06-12
CKT-2925: Recurring order models

krishhna123:
Adding new test-data models for `Recurring order` and `Counter`.
For now we are just adding graphql model and rest models will be added once recurring order is available in public beta


CarlosCortizasCT:
Please, take into account the change in [this PR](https://github.com/commercetools/test-data/pull/863) which will require a small adjustment in the `postBuild` callbacks we're introducing in this one.

.changeset/tender-snails-search.md
```diff
@@ -0,0 +1,22 @@
+---
```
CarlosCortizasCT, line 1: Could we include here a note about REST models not being implemented by now?

standalone/src/models/recurring-order/counter/builders.ts
```diff
@@ -0,0 +1,10 @@
+import { createSpecializedBuilder } from '@/core';
+import { graphqlFieldsConfig } from './fields-config';
+import type { TCreateCounterBuilder, TCounterGraphql } from './types';
+
+export const GraphqlModelBuilder: TCreateCounterBuilder<TCounterGraphql> = () =>
+  createSpecializedBuilder({
+    name: 'counterGraphqlBuilder',
```
CarlosCortizasCT, line 7:
```suggestion
    name: 'CounterGraphqlBuilder',
```

standalone/src/models/recurring-order/counter/counter-draft/builders.ts
```diff
@@ -0,0 +1,12 @@
+import { createSpecializedBuilder } from '@/core';
+import type { TCreateCounterBuilder, TCounterDraftGraphql } from '../types';
+import { graphqlFieldsConfig } from './fields-config';
+
+export const GraphqlModelBuilder: TCreateCounterBuilder<
+  TCounterDraftGraphql
+> = () =>
+  createSpecializedBuilder({
+    name: 'counterDraftGraphqlBuilder',
```
CarlosCortizasCT, line 9:
```suggestion
    name: 'CounterDraftGraphqlBuilder',
```

standalone/src/models/recurring-order/recurring-order/builders.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+        resumesAt: null,
+        expiresAt: null,
+        lastOrderAt: null,
+        nextOrderAt: null,
+        skipConfiguration: null,
+        businessUnit: expect.objectContaining({
+          id: expect.any(String),
+          unitType: 'Company',
+          __typename: 'BusinessUnit',
+        }),
```
CarlosCortizasCT, line 31: When validating nested models in a GraphQL object, using the `__typename` property should be enough.

standalone/src/models/recurring-order/recurring-order/constants.ts
```diff
@@ -0,0 +1,6 @@
+export const recurringOrderState = {
+  Active: 'Active',
+  Paused: 'Paused',
+  Expired: 'Expired',
+  Canceled: 'Canceled',
+} as const;
```
CarlosCortizasCT, line 6: I think we should be using the `enum` from the generated types. Otherwise, I think Typescript will throw an error if we try to assign a string to a typed enum value.
```suggestion
import { TCtpRecurringOrderState } from '@/graphql-types';

export const recurringOrderState = TCtpRecurringOrderState;
```

.changeset/tender-snails-search.md
krishhna123: Done

standalone/src/models/recurring-order/counter/builders.ts
krishhna123: Done

standalone/src/models/recurring-order/counter/counter-draft/builders.ts
krishhna123: Done

standalone/src/models/recurring-order/recurring-order/builders.spec.ts
krishhna123: I am just making sure that nested models are generated properly in `postBuild` callback by adding extra validation.

CarlosCortizasCT: If you prefer keeping it like this it's fine, but it's not something we will be doing everywhere.
If the `__typename` of the object matches the expectation it's because the `postBuild` generated the right one.

krishhna123: Lets follow how its done everywhere :-) Just validating `__typename` for nested models

PR #870, opened 2025-06-17
[FEC-51]: Adding cart sub-models

NickDevG:
Adding missing sub-models for carts:
- `Shipping`
- `ItemShippingDetails`
- `ItemShippingTarget`

PR #875, opened 2025-06-18
Implement Resource Identifier model

nima-ct:
Recurrence policies are still not available in the SDK and this PR introduces a `Resource Identifier` model that will be used to mock a "drafted" Recurrence Policy field inside a standalone price form. See https://next-docs-subscriptions-docs.commercetools.vercel.app/api/projects/standalone-prices#standalonepricedraft

`StandalonePriceDraft` builder needs some adjustments in our e2e test in order to mock properly a recurrence policy field builder.

### DIFF
--- a/.changeset/thirty-worlds-matter.md
+++ b/.changeset/thirty-worlds-matter.md
@@ -2,4 +2,22 @@
 '@commercetools/composable-commerce-test-data': minor
 ---
 
-Implement resource identifier model
+We're introducing a new model named `ResourceIdentifier` that can be consumed from the `@commercetools/composable-commerce-test-data/commons` entry point.
+
+This is how the new model could be used:
+
+```ts
+import { ResourceIdentifierRest } from '@commercetools/composable-commerce-test-data/commons';
+
+const restModel = ResourceIdentifierRest.random().build();
+```
+
+```ts
+import { ResourceIdentifierGraphql } from '@commercetools/composable-commerce-test-data/commons';
+
+const graphqlModel = ResourceIdentifierGraphql.random().build();
+```
+
+`ResourceIdentifier` model provide support for mocking reference-like fields for draft models, such as `StandalonePriceDraft`.
+
+This model allows setting either `id` or `key`, as expected in reference fields across draft types (e.g. `channel`, `customerGroup`, `recurrencePolicy`).
--- a/.changeset/thirty-worlds-matter.md
+++ b/.changeset/thirty-worlds-matter.md
@@ -1,0 +1,5 @@
+---
+'@commercetools/composable-commerce-test-data': minor
+---
+
+Implement resource identifier model
--- a/standalone/src/models/commons/resource-identifier/builders.spec.ts
+++ b/standalone/src/models/commons/resource-identifier/builders.spec.ts
@@ -29,13 +29,13 @@ const validateGraphqlModel = (graphqlModel: TResourceIdentifierGraphql) => {
 };
 
 describe('ResourceIdentifier model builders', () => {
-  it('builds a REST model', () => {
+  it('builds a REST model with default values', () => {
     const restModel = ResourceIdentifierRest.random().build();
 
     validateRestModel(restModel);
   });
 
-  it('builds a populated REST model', () => {
+  it('builds a REST model with custom values', () => {
     const restModel = ResourceIdentifierRest.random()
       .id('12345')
       .key('test-key')
@@ -47,13 +47,13 @@ describe('ResourceIdentifier model builders', () => {
     expect(restModel.typeId).toEqual('foo');
   });
 
-  it('builds a GraphQL model', () => {
+  it('builds a GraphQL model with default values', () => {
     const graphqlModel = ResourceIdentifierGraphql.random().build();
 
     validateGraphqlModel(graphqlModel);
   });
 
-  it('builds a populated GraphQL model', () => {
+  it('builds a GraphQL model with custom values', () => {
     const graphqlModel = ResourceIdentifierGraphql.random()
       .id('12345')
       .key('test-key')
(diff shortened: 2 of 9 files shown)


mario-priceless (approved):

standalone/src/models/commons/resource-identifier/builders.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+    expect(restModel.typeId).toEqual('foo');
+  });
+
+  it('builds a GraphQL model', () => {
+    const graphqlModel = ResourceIdentifierGraphql.random().build();
+
+    validateGraphqlModel(graphqlModel);
+  });
+
+  it('builds a populated GraphQL model', () => {
```
krishhna123, line 56: I think instead of adding multiple test to validate building of fields, you can validate default model with `typeId`. e.g. `ResourceIdentifierGraphql.random().typeId('foo')build();`

dogayuksel (commented):
isn't resource identifier just a fancy name for draft of reference?

https://next-docs-subscriptions-docs.commercetools.vercel.app/api/types#resourceidentifier

https://github.com/commercetools/test-data/blob/8ee9cda7715874ac49720cd891360d16b942cd2f/standalone/src/models/commons/reference/reference-draft/fields-config.ts#L4-L9

standalone/src/models/commons/resource-identifier/index.ts
```diff
@@ -0,0 +1,13 @@
+import { RestModelBuilder, GraphqlModelBuilder } from './builders';
+import * as ResourceIdentifierPresets from './presets';
+export * from './types';
```
krishhna123, line 3: Instead of exporting types here, you can export them at `commons/index.ts`. That's how we want to manage our exports and avoid nested exports

standalone/src/models/commons/resource-identifier/fields-config.ts
```diff
@@ -0,0 +1,25 @@
+import { fake, type TModelFieldsConfig } from '@/core';
+import type {
+  TResourceIdentifierGraphql,
+  TResourceIdentifierRest,
+} from './types';
+
+const commonFieldsConfig = {
+  key: fake((f) => f.string.uuid()),
+  id: fake((f) => f.string.uuid()),
```
dogayuksel, line 9: this might be a bit misleading, it is either id or (whenever supported) key.
to model that, it might be possible to drop key on post build whenever id exists.

.changeset/thirty-worlds-matter.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools/composable-commerce-test-data': minor
+---
+
+Implement resource identifier model
```
krishhna123, line 5: You can follow new format of adding `changeset`. You can also refer https://github.com/commercetools/test-data/pull/870

krishhna123:
Maybe I am thinking too far ahead, but as `Resource Identifier` as a Draft type to create `Reference` and `KeyReference`. We can get rid of `ReferenceDraft` and `KeyReferenceDraft` which are not actual entities either on REST or GraphQL afaik.
In that case we might need to migrate presets.

nima-ct:
@dogayuksel @krishhna123
`ResourceIdentifier` is the closest type I could find to represent a [RecurrencePolicyResourceIdentifier](https://next-docs-subscriptions-docs.commercetools.vercel.app/api/projects/recurrence-policies#recurrencepolicyresourceidentifier), which is not available yet in the SDK (still in beta)
It represents the type for the `recurrencePolicy` field of a StandalonePriceDraft
https://next-docs-subscriptions-docs.commercetools.vercel.app/api/projects/standalone-prices#standalonepricedraft

The main goal of this model was to align with the external types while still allowing access to the key field.
The need comes from the requirement to populate a dropdown using the `key` reference of a `recurrence policy`, similar to what’s done for the `channel` field here:
https://github.com/commercetools/merchant-center-prices/blob/c121ee1eafdf04343a9de3c7e0d8ee7455ac2430/cypress/support/commands.ts#L379

After the changes in my previous[PR](https://github.com/commercetools/test-data/pull/871), `StandalonePriceDraft` aligns more strictly with the SDK types. Using `Reference` or `KeyReference` model will be rejected by TypeScript when trying to build a `recurrencePolicy`

nima-ct:
> isn't resource identifier just a fancy name for draft of reference?
> 
> https://next-docs-subscriptions-docs.commercetools.vercel.app/api/types#resourceidentifier
> 
> https://github.com/commercetools/test-data/blob/8ee9cda7715874ac49720cd891360d16b942cd2f/standalone/src/models/commons/reference/reference-draft/fields-config.ts#L4-L9

For the recurrence policy field, I needed a `key` reference for the tests that could satisfy the `StandalonePriceDraft` type

.changeset/thirty-worlds-matter.md
```diff
@@ -0,0 +1,23 @@
+---
+'@commercetools/composable-commerce-test-data': minor
+---
+
+We're introducing a new model named `ResourceIdentifier` that can be consumed from the `@commercetools/composable-commerce-test-data/commons` entry point.
+
+This is how the new model could be used:
+
+```ts
```
krishhna123, line 9: nit:
```suggestion
```ts
import {
ResourceIdentifierRest, 
ResourceIdentifierGraphql }
 from '@commercetools/composable-commerce-test-data/commons';

const restModel = ResourceIdentifierRest.random().build();
const graphqlModel = ResourceIdentifierGraphql.random().build();
```
```

nima-ct:
@krishhna123 
Thanks for your review :raised_hands:
We realized that  `KeyReferenceDraft` already satisfies the type requirements for recurrencePolicy. It was initially rejected due to mixed imports from both the old and new test-data packages in the test file, which led to type conflicts.

I was importing `KeyReferenceDraft` from the legacy `@commercetools-test-data` and trying to use it inside a `StandalonePriceDraft` builder imported from the new `@commercetools/composable-commerce-test-data` 

We don’t have a use case for `ResourceIdentifier` anymore so unless you think it’s worth keeping around, I’m going to close this PR.

krishhna123:
> @krishhna123 Thanks for your review 🙌 We realized that `KeyReferenceDraft` already satisfies the type requirements for recurrencePolicy. It was initially rejected due to mixed imports from both the old and new test-data packages in the test file, which led to type conflicts.
> 
> I was importing `KeyReferenceDraft` from the legacy `@commercetools-test-data` and trying to use it inside a `StandalonePriceDraft` builder imported from the new `@commercetools/composable-commerce-test-data`
> 
> We don’t have a use case for `ResourceIdentifier` anymore so unless you think it’s worth keeping around, I’m going to close this PR.

Yes.. I knew that we already had `KeyReferenceDraft` and `ReferenceDraft` which are actually same as `ResourceIdentifier`.
But imho, this we should use this model instead of `KeyReferenceDraft` and `ReferenceDraft` I can not see these entities in api document and even as GraphQL types.
But this requires some efforts. Something for future.

Sorry to drag you @CarlosCortizasCT but I would love to hear your opinion.

PR #876, opened 2025-06-19
Refactor Order models directory layout

CarlosCortizasCT:
## Description

Since we will need to migrate the `Order` related test data models, I thought it would be helpful to restructure the directory layout in isolation so the upcoming work would be more focused.

The main change is that the `Order` model files were directly placed within the `/models/order` directory alongside the other submodels.
Now the `Order` models and its drafts ones have been moved to the `/model/order/order` directory.

I've also updated the way of exporting each model in a standard way which will help with the upcoming migration work.


standalone/src/models/payment/payment/payment-draft/presets/sample-data-b2b/lifttech-solutions-ltd-fr-payment-2-payment.ts
```diff
@@ -19,10 +13,6 @@ const camilleLefevre = CustomerDraft.presets.sampleDataB2B
   .camilleLefevre()
   .build<TCustomerDraft>();
 
-const order = OrderFromCartDraft.presets.sampleDataB2B
```
CarlosCortizasCT, line 22: This was not used in the file.

PR #878, opened 2025-06-20
[FEC-51]: migrating order models #1 - TrackingData

NickDevG:
Starting to migrate order models staring with: TrackingData


standalone/src/models/order/parcel/builder.spec.ts
```diff
@@ -68,6 +68,8 @@ describe('builder', () => {
       .id(id)
       .key(key)
       .measurements(measurements)
+      // Parcel will be migrated and this file won't exist anymore
+      // @ts-expect-error - TrackingData is not a valid type for Parcel
```
NickDevG, line 72: The idea is to migrate `Parcel`, for that I need to migrate `TrackingData`, `DeliveryItem` and `ParcelMeasurements`. 

This file will be removed shortly

.changeset/easy-worms-sell.md
```diff
@@ -0,0 +1,13 @@
+---
+'@commercetools/composable-commerce-test-data': minor
+---
+
+We're introducing a new model named `TrackingDataGraphql` that can be consumed from the `@commercetools/composable-commerce-test-data/order` entry point.
```
CarlosCortizasCT, line 5: I think this can be misleading as we're not introducing a new model but migrating an existing one which still exports a GraphQL, REST and compat versions.

standalone/src/models/order/tracking-data/presets/index.ts
```diff
@@ -0,0 +1,3 @@
+const presets = {};
```
CarlosCortizasCT, line 1: We should export the three types of presets for consistency:
```
export const compatPresets = {};
```

This makes sure if we need to add a preset in the future, the structure is already there for whoever needs to do it.

standalone/src/models/order/tracking-data/index.ts
```diff
@@ -1,5 +1,24 @@
...
+import {
+  RestModelBuilder,
+  GraphqlModelBuilder,
+  CompatTrackingDataModelBuilder,
+} from './builders';
+import * as modelPresets from './presets';
 
+export const TrackingDataRest = {
+  random: RestModelBuilder,
+  presets: modelPresets.default,
```
CarlosCortizasCT, line 10: We should be using `modelsPresets.restPresets`.

CarlosCortizasCT (approved):
Thanks! 👍

PR #879, opened 2025-06-20
[FEC-51]: migrating order models #2 - DeliveryItem

NickDevG:
Continue to migrate order models, now: `DeliveryItem`

### DIFF
--- a/.changeset/yummy-owls-cough.md
+++ b/.changeset/yummy-owls-cough.md
@@ -2,9 +2,9 @@
 '@commercetools/composable-commerce-test-data': minor
 ---
 
-We're introducing a new model named `DeliveryItemGraphql` that can be consumed from the `@commercetools/composable-commerce-test-data/order` entry point.
+We're migrating the model `DeliveryItem` that can be now consumed from the `@commercetools/composable-commerce-test-data/order` entry point.
 
-This is how the new models could be used:
+This is how the migrated model could be used:
 
 ```ts
 import { DeliveryItemGraphql } from '@commercetools/composable-commerce-test-data/order';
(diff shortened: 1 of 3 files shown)


standalone/src/models/order/delivery-item/presets/index.ts
```diff
@@ -0,0 +1,3 @@
+const presets = {};
+
+export default presets;
```
CarlosCortizasCT, line 3: We should export the three types of presets for consistency:
```
export const compatPresets = {};
```

This makes sure if we need to add a preset in the future, the structure is already there for whoever needs to do it.

standalone/src/models/order/delivery-item/index.ts
```diff
@@ -1,5 +1,22 @@
-import LegacyBuilder from './builder';
+import {
+  RestModelBuilder,
+  GraphqlModelBuilder,
+  CompatDeliveryItemModelBuilder,
+} from './builders';
+import * as modelPresets from './presets';
```
CarlosCortizasCT, line 6: We should be including the corresponding `preset` attribute in each model.

.changeset/yummy-owls-cough.md
```diff
@@ -0,0 +1,13 @@
+---
+'@commercetools/composable-commerce-test-data': minor
+---
+
+We're introducing a new model named `DeliveryItemGraphql` that can be consumed from the `@commercetools/composable-commerce-test-data/order` entry point.
```
CarlosCortizasCT, line 5: I think this can be misleading as we're not introducing a new model but migrating an existing one which still exports a GraphQL, REST and compat versions.

CarlosCortizasCT (approved):
Thanks! 👍

krishhna123 (approved):

PR #881, opened 2025-06-20
Fix CategoryDraft model

CarlosCortizasCT:
## Description

We're fixing the `CategoryDraft` model as it was populating the `slug` field incorrectly by default.
It was including three slugs in three different languages (`de`, `en` and `fr`) but two of them were including spaces in the values, which is not allowed.

### BEFORE
Screenshot 2025-06-20 at 17 54 14

### AFTER
Screenshot 2025-06-20 at 17 54 23

### DIFF
--- a/.changeset/old-cobras-attack.md
+++ b/.changeset/old-cobras-attack.md
@@ -1,0 +1,6 @@
+---
+'@commercetools/composable-commerce-test-data': patch
+---
+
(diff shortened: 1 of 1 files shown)


krishhna123 (approved):
It looks fine.

NickDevG (approved):

PR #883, opened 2025-06-25
[FEC-51]: migrating order models #4 - Parcels

NickDevG:
Continue to migrate order models, now: Parcels


CarlosCortizasCT (approved):
Looks good to me 👍 

Thanks!

PR #884, opened 2025-06-25
Remove sample data presets - batch 1

CarlosCortizasCT:
## Description

This PR starts the work for removing the sample data presets from all the test-data models in favor of the new `@commercetools/composable-commerce-test-data-sample-presets` package.

Since there are thousands of file, we'll be creating several PRs to split the reviews.


CarlosCortizasCT:
Need to better coordinate the sample data presets to be removed.

PR #891, opened 2025-06-26
[FEC-51]: migrating order models #5 - LineItemReturnItem

NickDevG:
Continue to migrate order models, now: `LineItemReturnItem`


standalone/src/models/order/line-item-return-item/index.ts
```diff
@@ -1,7 +1,26 @@
-import LegacyBuilder from './builder';
-import * as constants from './constants';
+import {
+  RestModelBuilder,
+  GraphqlModelBuilder,
+  CompatLineItemReturnItemModelBuilder,
+} from './builders';
+import * as modelPresets from './presets';
 
+export * as constants from './constants';
```
CarlosCortizasCT, line 8: Constants should be exported as part of each model ([example](https://github.com/commercetools/test-data/blob/main/standalone/src/models/product-type/attribute-definition/index.ts)).

CarlosCortizasCT: We will also need to update the constants to export the TS enum so we don't have TS typing issues with the GraphQL model.
[Here's a reference](https://github.com/commercetools/test-data/pull/864#discussion_r2150321232) of what we've done in another model.

image

image

standalone/src/models/order/line-item-return-item/constants.ts
```diff
@@ -1,15 +1,7 @@
...
-};
+import { TCtpShipmentState, TCtpPaymentState } from '@/graphql-types';
 
-export const paymentState = {
-  nonRefundable: 'NonRefundable',
-  initial: 'Initial',
-  refunded: 'Refunded',
-  notRefunded: 'NotRefunded',
-};
+export const shipmentState = TCtpShipmentState;
```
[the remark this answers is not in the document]
NickDevG: Nice catches @CarlosCortizasCT. I saw one match and assumed it was the right enum. Thanks!

Sarah4VT (approved):
Looks good! This one was a good easy first one to understand after reading the docs.

standalone/src/models/order/line-item-return-item/builders.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+const validateGraphqlFields = (model: TLineItemReturnItemGraphql) => {
+  validateCommonFields(model);
+  expect(model).toEqual(
+    expect.objectContaining({
+      __typename: 'LineItemReturnItem',
+    })
+  );
+};
+
+describe('LineItemReturnItem model builders', () => {
```
Sarah4VT, line 40: Noticed this after I approved. Same as comment in the next PR.

The docs state "It's also recommended to test a customized version of the data models as well:". I think this is where this test would go but don't see it. Thought I'd mention it.

CarlosCortizasCT: That's a recommendation we're not following so far so I wouldn't block all these PRs because of it.
It is something that can be done in the future if we find it valuable.

PR #892, opened 2025-06-26
[FEC-51]: migrating order models #6 - ReturnInfo

NickDevG:
Continue to migrate order models, now: `ReturnInfo`


standalone/src/models/order/return-info/fields-config.ts
```diff
@@ -0,0 +1,33 @@
...
+  returnTrackingId: fake((f) => f.string.uuid()),
+  returnDate: fake(returnDate),
+};
+
+export const restFieldsConfig: TModelFieldsConfig<TReturnInfoRest> = {
+  fields: {
+    ...commonFieldsConfig,
+    items: fake((f) =>
+      f.helpers.arrayElement([LineItemReturnItemRest.random()])
+    ),
```
CarlosCortizasCT, line 21: `arrayElement` picks one random item from an array but here we just have one item so it doesn't seem its needed.

```suggestion
    items: fake((f) =>[LineItemReturnItemRest.random()]),
```

standalone/src/models/order/return-info/fields-config.ts
```diff
@@ -0,0 +1,33 @@
...
+  },
+};
+
+export const graphqlFieldsConfig: TModelFieldsConfig<TReturnInfoGraphql> = {
+  fields: {
+    ...commonFieldsConfig,
+    items: fake((f) =>
+      f.helpers.arrayElement([LineItemReturnItemGraphql.random()])
+    ),
```
CarlosCortizasCT, line 30:
```suggestion
    items: fake((f) => [LineItemReturnItemGraphql.random()]),
```

CarlosCortizasCT (approved):
Thanks! 👍

PR #898, opened 2025-06-30
Fix CustomFieldsRest test data model

CarlosCortizasCT:
## Description

Fixed `CustomFieldsRest` test data model as the default generated version was not populating the `fields` property but that's a required property.

Now we populate it with an empty object.

### DIFF
--- a/standalone/src/models/custom-fields/custom-fields/builders.spec.ts
+++ b/standalone/src/models/custom-fields/custom-fields/builders.spec.ts
@@ -31,6 +31,7 @@ const validateRestModel = (model: TCustomFieldsRest) => {
       fields: expect.any(Object),
     })
   );
+  expect(model.fields).not.toBeNull();
 };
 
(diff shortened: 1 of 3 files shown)


standalone/src/models/custom-fields/custom-fields/builders.spec.ts
```diff
@@ -31,44 +31,54 @@ const validateRestModel = (model: TCustomFieldsRest) => {
       fields: expect.any(Object),
     })
   );
+  expect(model.fields).not.toBeNull();
```
Sarah4VT, line 34: Is there a reason you validated this field is not null in REST but for GraphQL you did it as an additional test?

Sarah4VT (approved):

standalone/src/models/custom-fields/custom-fields/builders.spec.ts
CarlosCortizasCT: This field does not exist in the GraphQL version of the model.

PR #902, opened 2025-07-01
New ItemState test data model

CarlosCortizasCT:
## Description

We're introducing a new test data model named `ItemState` which can be consumed from the `@commercetools/composable-commerce-test-data/cart` entry point.

[Here](https://docs.commercetools.com/api/projects/carts#itemstate) are the official docs for this resource.

This is how it can be used:

```ts
import {
  ItemStateGraphql,
  ItemStateRest,
} from '@commercetools/composable-commerce-test-data/cart';

const graphqlModel = ItemStateGraphql.random().build();
const restModel = ItemStateRest.random().build();
```

### DIFF
--- a/.changeset/orange-results-hang.md
+++ b/.changeset/orange-results-hang.md
@@ -1,0 +1,17 @@
+---
+'@commercetools/composable-commerce-test-data': patch
+---
+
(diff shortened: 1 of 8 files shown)


bradd123 (approved):
💯

PR #903, opened 2025-07-01
New CartDiscountTarget test data model

CarlosCortizasCT:
## Description

We're introducing a new test data model named `CartDiscountTarget` which can be consumed from the `@commercetools/composable-commerce-test-data/cart-discount` entry point.

[Here](https://docs.commercetools.com/api/projects/cartDiscounts#cartdiscounttarget) are the official docs for this resource.

This is how it can be used:

```ts
import {
  CartDiscountTargetGraphql,
  CartDiscountTargetRest,
} from '@commercetools/composable-commerce-test-data/cart-discount';

const graphqlModel = CartDiscountTargetGraphql.random()
  .type(CartDiscountTargetGraphql.constants.targetTypes.lineItems)
  .build();

const restModel = CartDiscountTargetRest.random()
  .type(CartDiscountTargetGraphql.constants.targetTypes.shipping)
  .build();
```

### DIFF
--- a/standalone/src/models/cart/cart-discount/cart-discount-target/fields-config.ts
+++ b/standalone/src/models/cart/cart-discount/cart-discount-target/fields-config.ts
@@ -1,0 +1,46 @@
+import { fake, type TModelFieldsConfig } from '@/core';
+import { targetTypes } from './constants';
+import type {
+  TCartDiscountTargetGraphql,
+  TCartDiscountTargetRest,
+} from './types';
+
+// https://docs.commercetools.com/api/projects/cartDiscounts#cartdiscounttarget
+
+const targetTypeToGraphqlTypenameMap: Record<string, string> = {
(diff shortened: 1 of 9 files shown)


standalone/src/models/cart/cart-discount/cart-discount-target/fields-config.ts
```diff
@@ -0,0 +1,46 @@
...
+};
+
+export const graphqlFieldsConfig: TModelFieldsConfig<TCartDiscountTargetGraphql> =
+  {
+    fields: {
+      ...commonFieldsConfig,
+      predicate: null,
+      __typename: null,
+    },
+    postBuild: (model) => {
```
CarlosCortizasCT, line 35: I tried to replicate what we already have in the monorepo ([reference](https://github.com/commercetools/merchant-center-frontend/blob/main/packages-shared/test-data/src/cart-discount-target/transformers.js)).

PR #906, opened 2025-07-01
Fec 45 - migrate State model

Sarah4VT:
The State test data model was missing some presets used in mc-fe so I added those here and migrated the model to the new implementation patterns.


CarlosCortizasCT:
There were some things that I thought need some adjustments so I took the time to push some changes to the PR:
1. We usually keep the `constants` file and export its contents through the exported models so consumers can use the typed valued easier (related commit: 7d8c6ad6c32220f9e63198be0281c198a9105e5c)
2. We usually keep just one builder type for each model which we reuse for both the read and write models (related commit: f6f8c72a9dbb582d75e34cbfbac087e0b7e6e56b)
3. In the models' fields configuration we only populate the mandatory fields. There was also some adjustment needed for the compatibility builder. We addressed the read model changes. (related commit: 943006e703a99a98f5401742e16474116cd41a68)
4. We also adjusted the default values for the write model (related commit: a6a74c5f5977b2166f87ac8c874a15b86e670757)
5. Give a bit more context in the changeset file (related commit: b77918ea5faeb584690b266afdd7074b1e38e978)

PR #907, opened 2025-07-02
Add new presets for Cart related test data models

CarlosCortizasCT:
### Additions

We're introducing some presets in `Cart` related test data models which should help consumers building objects.

#### LineItem

New `withAllFields` preset which populates all the model properties.
Example:

```ts
import { LineItemGraphql } from '@commercetools/composable-commerce-test-data/cart';

const fullModel = LineItemGraphql.presets.withAllFields();
```

#### TaxedPrice

New `withCurrency` preset which populates the `totalNet`, `totalGross` and `taxPortions` properties with `Money` objects using a specific currency code. `EUR` is used by default.
Example:

```ts
import { TaxedItemPrice } from '@commercetools/composable-commerce-test-data/cart';

const fullModel = TaxedItemPrice.presets.withAllFields();
```

#### TaxedItemPrice

We're updating the existing `withAllFields` preset in the `TaxedItemPrice` test data model.
This preset was populating the `totalNet`, `totalGross`, `totalTax` and `taxPortions` with `Money` objects using always `EUR` as the `currencyCode`.
Now the preset allows consumers to provide a different currency code using `EUR` as its default.

### DiscountedLineItemPriceForQuantity

We're adjusting the way the `quantity` property is populated. This property represents "_Number of Line Items or Custom Line Items in the Cart_" (https://docs.commercetools.com/api/projects/carts#discountedlineitempriceforquantity) but we were using any integer value which was yielding values like 12342423, which does not make sense.
Now we populate this property by default with a number between 1 and 10.


standalone/src/models/cart/cart/discounted-line-item-price-for-quantity/fields-config.ts
```diff
@@ -9,7 +9,7 @@ import type {
 } from './types';
 
 const commonFieldsConfig = {
-  quantity: fake((f) => f.number.int()),
+  quantity: fake((f) => f.number.int({ min: 1, max: 10 })),
```
CarlosCortizasCT, line 12: The number here represents the amount of items in the cart with discounts and we were using very high numbers for it.

standalone/src/models/cart/cart/line-item/index.ts
```diff
@@ -7,22 +7,20 @@ import * as constants from './constants';
 import * as modelPresets from './presets';
 
 export const LineItemRest = {
+  constants,
```
CarlosCortizasCT, line 10: We need to export the constants in every model.

Sarah4VT (approved):

PR #912, opened 2025-07-04
New OAuthClient test data model

CarlosCortizasCT:
## Description

Three new models have been added related to MC Gateway `OAuthClient` entity:

- `OAuthClient`
- `OAuthScope`
- `ProjectPermission`

They all can be consumed from this new entry point: `@commercetools/composable-commerce-test-data/oauth-client`.

They only have a GraphQL representation as the MC Gateway only exposes a GraphQL API.

Here's an example of how to use them:

```ts
import {
  OAuthClientGraphql,
  OAuthScopeGraphql,
  ProjectPermission,
} from '@commercetools/composable-commerce-test-data/oauth-client';

const aouthClientModel = OAuthClientGraphql.random().permissions([
  ProjectPermissionGraphql.random().projectKey('my-project-key'),
]);

const oAuthScopeModel = OAuthScopeGraphql.random().name(
  OAuthScopeGraphql.constants.presets.viewCategories
);
```


standalone/package.json
```diff
@@ -83,8 +88,8 @@
       "./core/index.ts",
       "./core/test-utils.ts",
       "./custom-application.ts",
-      "./custom-object.ts",
       "./custom-fields.ts",
+      "./custom-object.ts",
```
CarlosCortizasCT, line 92: This was just to keep the alphabetical order.

.changeset/famous-deer-scream.md
```diff
@@ -0,0 +1,31 @@
...
+Here's an example of how to use them:
+
+```ts
+import {
+  OAuthClientGraphql,
+  OAuthScopeGraphql,
+  ProjectPermission,
+} from '@commercetools/composable-commerce-test-data/oauth-client';
+
+const aouthClientModel = OAuthClientGraphql.random().permissions([
```
Sarah4VT, line 24:
```suggestion
const oAuthClientModel = OAuthClientGraphql.random().permissions([
```

standalone/src/models/oauth-client/oauth-scope/constants.ts
```diff
@@ -0,0 +1,49 @@
+export const scopes = {
```
ByronDWall, line 1: Is this meant to be a comprehensive list of available scopes? 

If so, when new scopes are added to the api, do they need to be added here as well? 

Can the scopes be generated, or does this need to be updated manually?  e.g. Is this list of scopes the output of [the getEnabledProjectBoundScopes method](https://github.com/commercetools/merchant-center-services/blob/main/packages/gateway/lib/graphql/oauth-clients/helpers.ts#L147) in the gateway's oauth client helpers?

ByronDWall (approved):
Looks good to me, just one nit about potentially commenting the scopes constant that doesnt affect approval

Sarah4VT (approved):
One name that I think had a typo, but other than that, I didn't see anything.

standalone/src/models/oauth-client/oauth-scope/constants.ts
Sarah4VT: I also was wondering if these values could come from a generated list somewhere to keep the maintenance low.

CarlosCortizasCT: No, I don't think this needs to be a comprehensive list.
I brought here the list we already had [in the monorepo test data model](https://github.com/commercetools/merchant-center-frontend/blob/main/packages-shared/test-data/src/oauth-client/constants.js) which seems to have served well so far and that's why I think it's good enough.

.changeset/famous-deer-scream.md
CarlosCortizasCT: Thanks!

It has been updated here: 47a6a522386f402c049f3554f86a5e5b2158d682

PR #920, opened 2025-07-09
[Shopping list]: migrating the models

NickDevG:
This PR migrates the existing models for shopping list and shopping list draft to the new pattern while adding the (missing) sub-models shopping list line item and text line item.

The existing (old) models were using more than required fields, in case a consumer (that I couldn't find) needs that I'm adding a `with-all-fields` preset.

### DIFF
--- a/standalone/src/models/shopping-list/text-line-item/text-line-item-draft/builders.ts
+++ b/standalone/src/models/shopping-list/text-line-item/text-line-item-draft/builders.ts
@@ -10,7 +10,7 @@ export const RestModelBuilder: TCreateTextLineItemBuilder<
   TTextLineItemDraftRest
 > = () =>
   createSpecializedBuilder({
-    name: 'TextLineItemRestBuilder',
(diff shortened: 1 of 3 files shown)


ChristianMoll (approved):

standalone/src/models/shopping-list/text-line-item/text-line-item-draft/builders.ts
```diff
@@ -0,0 +1,25 @@
...
+    name: 'TextLineItemRestBuilder',
+    type: 'rest',
+    modelFieldsConfig: restFieldsConfig,
+  });
+
+export const GraphqlModelBuilder: TCreateTextLineItemBuilder<
+  TTextLineItemDraftGraphql
+> = () =>
+  createSpecializedBuilder({
+    name: 'TextLineItemGraphqlBuilder',
```
leventekobor, line 22: name should be `TextLineItemDraftGraphqlBuilder` right?

PR #921, opened 2025-07-10
Migrate AssociateRole test data model

CarlosCortizasCT:
## Description

We've migrated the `AssociateRole` model to the new implementation patterns.

The default object generated by this model was populating the `name` property which is not correct as that's an optional field so not the model won't populate it.

### DIFF
--- a/standalone/src/models/associate-role/fields-config.ts
+++ b/standalone/src/models/associate-role/fields-config.ts
@@ -5,17 +5,6 @@ import type { TAssociateRoleGraphql, TAssociateRoleRest } from './types';
 
 const [getOlderDate, getNewerDate] = createRelatedDates();
 
-// buyerAssignable: Scalars['Boolean']['output'];
-// createdAt: Scalars['DateTime']['output'];
-// createdBy?: Maybe<TCtpInitiator>;
-// custom?: Maybe<TCtpCustomFieldsType>;
-// id: Scalars['String']['output'];
-// key: Scalars['String']['output'];
-// lastModifiedAt: Scalars['DateTime']['output'];
-// lastModifiedBy?: Maybe<TCtpInitiator>;
-// name?: Maybe<Scalars['String']['output']>;
-// permissions: Array<TCtpPermission>;
-// version: Scalars['Long']['output'];
 const commonFieldsConfig = {
   buyerAssignable: fake((f) => f.datatype.boolean()),
   createdAt: fake(getOlderDate),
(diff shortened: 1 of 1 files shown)


standalone/src/models/associate-role/fields-config.ts
```diff
@@ -0,0 +1,67 @@
+import { fake, sequence, type TModelFieldsConfig } from '@/core';
+import { ClientLogging } from '@/models/commons';
+import { createRelatedDates } from '@/utils';
+import type { TAssociateRoleGraphql, TAssociateRoleRest } from './types';
+
+const [getOlderDate, getNewerDate] = createRelatedDates();
+
+// buyerAssignable: Scalars['Boolean']['output'];
```
ByronDWall, line 8: I really like adding the expected types as a comment, it makes understanding what's required here more easy to understand at-a-glance

standalone/src/models/associate-role/associate-role-draft/presets/empty.ts
```diff
@@ -1,11 +1,28 @@
...
-    .name(undefined)
-    .buyerAssignable(undefined)
-    .permissions(undefined)
-    .custom(undefined);
+const populatePreset = <
+  TModel extends TAssociateRoleDraftGraphql | TAssociateRoleDraftRest,
+>(
+  builder: TBuilder<TModel>
+) => {
+  return builder.name(null).permissions([]).custom(null);
```
ByronDWall, line 18: I realize that you are updating this preset to return values that are not optional in the `Draft` types, but should `buyerAssignable` be set to `null` for backwards compat, or is part of this process updating test files that consume this preset if/when necessary once this package is released?

standalone/src/models/associate-role/associate-role-draft/presets/empty.spec.ts
```diff
@@ -1,12 +1,52 @@
...
+import { restPreset, graphqlPreset, compatPreset } from './empty';
+
+const validateModel = (
+  model: TAssociateRoleDraftRest | TAssociateRoleDraftGraphql
+) => {
+  expect(model).toMatchObject({
+    key: expect.any(String),
+    name: null,
+    buyerAssignable: expect.any(Boolean),
+    permissions: expect.any(Array),
```
ByronDWall, line 14: Should this be `expect.toBe([])`?

CarlosCortizasCT: Update here: 9bad310a3cf270e2d46d1c4824d6e820ff87220a

standalone/src/models/associate-role/associate-role-draft/presets/empty.ts
CarlosCortizasCT: That's a good question/

I thought it doesn't make sense to set to `null` required values so I updated the preset.

I also tested this changes in my local MC frontend repo and nothing failed.

PR #922, opened 2025-07-10
Update test data models migration guidelines

CarlosCortizasCT:
## Description

I've took the time to migrate a new test-data model following the current guidelines and I found some areas which could be improved.

Also, I found the `Channel` model not following the expected pattern in the way it exports the models so I also adjusted it.


docs/guidelines/migrating-existing-model.md
Sarah4VT, line 1: This should help the AI prompts I've been using with some of the things it's been confused about. I also realized I've been using Channel as an example a lot because it was referenced here, so it's good to get this cleaned up!

Sarah4VT (approved):

ByronDWall (approved):
These updates are definitely helpful, thanks!

bradd123 (approved):

PR #923, opened 2025-07-10
feat(Discounts): add new DiscountGroup and DiscountGroupDraft models

ragafus:
This PR adds DiscountGroup and DiscountGroupDraft models

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -33,7 +33,7 @@
 /standalone/src/models/custom-view @commercetools/bots-team-engineers
 /standalone/src/models/customer @commercetools/customers-team-fe
 /standalone/src/models/customer-group @commercetools/customers-team-fe
-/standalone/src/models/discount-code @commercetools/priceless-team-fe
+/standalone/src/models/discount @commercetools/priceless-team-fe
 /standalone/src/models/discounts-custom-view @commercetools/priceless-team-fe
 /standalone/src/models/filter-values @commercetools/priceless-team-fe
 /standalone/src/models/inventory-entry @commercetools/pdm-team-fe


standalone/src/models/discount-group/discount-group-draft/types.ts
```diff
@@ -0,0 +1,10 @@
+import { DiscountGroupDraft } from '@commercetools/platform-sdk';
```
CarlosCortizasCT, line 1: We should remove this file as the draft model types already exist in the parent folder `types.ts` file.

standalone/src/models/discount-group/discount-group-draft/builders.ts
```diff
@@ -0,0 +1,25 @@
+import { createSpecializedBuilder } from '@/core';
+import { restFieldsConfig, graphqlFieldsConfig } from './fields-config';
+import type {
+  TCreateDiscountGroupDraftBuilder,
+  TDiscountGroupDraftGraphql,
+  TDiscountGroupDraftRest,
+} from './types';
```
CarlosCortizasCT, line 7: We should be using the types from the parent folder `type.ts` file.

standalone/src/models/discount-group/fields-config.ts
```diff
@@ -0,0 +1,38 @@
...
+  sortOrder: fake((f) =>
+    String(f.number.float({ min: 0.00000001, max: 0.99999999 }))
+  ),
+};
+
+export const restFieldsConfig: TModelFieldsConfig<TDiscountGroupRest> = {
+  fields: {
+    ...commonFieldsConfig,
+    name: fake(() => LocalizedString.random()),
+    description: fake(() => LocalizedString.random()),
```
CarlosCortizasCT, line 25: We are following the convention of only populating the required fields by default.

I know, we must document this.

standalone/src/models/discount-group/fields-config.ts
```diff
@@ -0,0 +1,38 @@
...
+
+export const graphqlFieldsConfig: TModelFieldsConfig<TDiscountGroupGraphql> = {
+  fields: {
+    ...commonFieldsConfig,
+    name: fake((f) => f.lorem.words(3)),
+    description: fake((f) => f.lorem.words(5)),
+    nameAllLocales: fake(() => LocalizedString.random()),
+    descriptionAllLocales: fake(() => LocalizedString.random()),
+    __typename: 'DiscountGroup',
+  },
```
CarlosCortizasCT, line 37: For GraphQL models, we compute the `name` and `description` fields out of the `nameAllLocales` and `descriptionAllLocales`.
Here's an example on how it would look like:
```ts
postBuild: (model) => {
    const name = model.nameAllLocales
      ? LocalizedString.resolveGraphqlDefaultLocaleValue(model.nameAllLocales)
      : null;

    const description = model.descriptionAllLocales
      ? LocalizedString.resolveGraphqlDefaultLocaleValue(
          model.descriptionAllLocales
        )
      : null;

    return {
      ...model,
      name,
      description,
    };
  },
  ```

File: standalone/src/models/discount-group/builders.spec.ts [hunk 5]
```diff
@@ -0,0 +1,73 @@
...
+    const graphqlModel = DiscountGroupGraphql.random().build();
+
+    expect(graphqlModel).toEqual(
+      expect.objectContaining({
+        id: expect.any(String),
+        createdAt: expect.any(String),
+        createdBy: expect.objectContaining({
+          clientId: expect.any(String),
+          externalUserId: expect.any(String),
+        }),
```
CarlosCortizasCT, line 44: For nested models in the GraphQL builder it's fine to just check for the `__typename` property. If it says it's the right type, we can trust on that model's tests that the value will be correctly generated.

```suggestion
        createdBy: expect.objectContaining({
          __typename: 'Initiator',
        }),
```

(_we will document this_)

standalone/src/models/discount-group/fields-config.ts
```diff
@@ -0,0 +1,38 @@
...
+};
+
+export const graphqlFieldsConfig: TModelFieldsConfig<TDiscountGroupGraphql> = {
+  fields: {
+    ...commonFieldsConfig,
+    name: fake((f) => f.lorem.words(3)),
+    description: fake((f) => f.lorem.words(5)),
+    nameAllLocales: fake(() => LocalizedString.random()),
+    descriptionAllLocales: fake(() => LocalizedString.random()),
```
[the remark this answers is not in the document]
ragafus: > We are following the convention of only populating the required fields by default.

Which value do you use for optional values? `Field` type does not allow undefined 🤔

CarlosCortizasCT:
@ragafus fyi I've opened [this PR](https://github.com/commercetools/test-data/pull/926) where I'm trying to improve the docs based on this PR's review

ragafus:
> One last suggestion: since we've already worked on trying to group a bit the implementation of related models and we already had a directory for the `discount-codes` model, would you consider grouping that one and the new one together in a new `discounts` folder?
> 
> We would keep the new entry point but we would just group those two models together:
> 
> ```
> /standalone/src/models/discounts/
>   /discount-code
>   /discount-group
> ```



> One last suggestion: since we've already worked on trying to group a bit the implementation of related models and we already had a directory for the `discount-codes` model, would you consider grouping that one and the new one together in a new `discounts` folder?
> 
> We would keep the new entry point but we would just group those two models together:
> 
> ```
> /standalone/src/models/discounts/
>   /discount-code
>   /discount-group
> ```

👍

What's the rule fo grouping models? To group `discount-code` and `discount-code` makes sense in my mental model because the owners(domain) is the same. But I was surprised when I realized that `cart-discount` (a type of discount) is grouped with `cart` related models 🤔

CarlosCortizasCT (approved):
Many thanks for addressing all the suggestions 🙇

PR #924, opened 2025-07-10
fix: wrong builder name when generating draft models

ragafus:
I just run `pnpm generate-model` to create some models and I realized that the name of the builder type for draft  models is wrong.
Is 
```ts
import type {
  TCreateDiscountGroupDraftBuilder,
  TDiscountGroupDraftGraphql,
  TDiscountGroupDraftRest,
} from '../types';
```
and should be
```ts
import type {
  TCreateDiscountGroupBuilder,
  TDiscountGroupDraftGraphql,
  TDiscountGroupDraftRest,
} from '../types';
```

I've introduced a new property (`nonDraftModelName`) when running `renderTemplate` for draft models.
Let me know if any other approach suits best with repo guidelines.

### DIFF
--- a/generators/src/new-test-model/new-test-model.generator.ts
+++ b/generators/src/new-test-model/new-test-model.generator.ts
@@ -176,7 +176,7 @@ export const newTestModelGenerator: CodeGenerator = {
             isPresetExampleRequired,
             modelName: `${modelName}Draft`,
             modelCodename: `${modelCodename}-draft`,
-            nonDraftModelName: modelName,
+            baseModelName: modelName,
             graphqlTypePrefix,
           }),
           { encoding: 'utf-8' }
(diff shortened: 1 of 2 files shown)


generators/src/new-test-model/new-test-model.generator.ts
```diff
@@ -176,6 +176,7 @@ export const newTestModelGenerator: CodeGenerator = {
             isPresetExampleRequired,
             modelName: `${modelName}Draft`,
             modelCodename: `${modelCodename}-draft`,
+            nonDraftModelName: modelName,
```
CarlosCortizasCT, line 179: The solution looks good to me.

I would only suggest to rename this property to `baseModelName`.

CarlosCortizasCT (approved):
Many thanks 👍

PR #927, opened 2025-07-11
fix(zone/location): migrate the Location test-data model (under zone) to the v2 builder pattern

Sarah4VT:
Migrated the `Location` test-data model (under `zone`) to the **v2 builder pattern**.

Highlights:
• Introduced separate REST (`LocationRest`) and GraphQL (`LocationGraphql`) builders plus a compatibility builder (`Location`) for full backwards-compatibility.
• Added representation-specific field configs; required `country` now always generated, optional `state` defaults to `null`.
• Converted presets and tests to `restPreset`, `graphqlPreset`, and `compatPreset` structure.

This is a backward-compatible change; existing `Location.random()` usages keep working, but consumers are encouraged to adopt the new explicit builders.


docs/guidelines/migrating-existing-model.md
Sarah4VT, line 1: Not specific to the model migration, but just a little docs cleanup.

standalone/src/models/store/store/store-draft/builders.ts
Sarah4VT, line 1: Just a **very** minor naming inconsistency.

standalone/src/models/zone/location/presets/change-history-data/with-country-no-state.ts
Sarah4VT, line 1: Removed this one as now this is the default. There may need to be a new general preset added if any tests are depending on having both values set and we find that is a pattern that is heavily used, but can add that when the need arises.

CarlosCortizasCT: If it's only used in one application, it should not live here.

CarlosCortizasCT (approved):
Looks good to me 💯 

Thanks!

bradd123 (approved):

PR #928, opened 2025-07-11
feat(ai): add some sample ai prompts

Sarah4VT:
This pull request introduces two new prompts to assist with migrating and verifying test-data models in the `commercetools-test-data` repository. These prompts aim to streamline the migration from the v1 to the v2 model pattern and ensure adherence to conventions for v2 models. The idea is that as developers use the model and discover areas where it may create incorrect or inconsistent results, the prompts can be refined to produce more consistent and reliable results. The hope is that this can speed up the developer time it takes to migrate models, which involves a fair amount of repetitive and pattern based changes that can be done much faster with AI.

Below are the prompts included with descriptions of what they do. How to use the prompts is outlined at the top of each prompt file. One thing to note is I do have a local rules file for my repository that could also be influencing results. I tried to take important pieces of it and include it in these prompts, but if results are inconsistent, we can always tweak these or add more.

### Migration Prompt (`migrate-v1-to-v2-model.md`)

* Added a detailed step-by-step guide for migrating v1 test-data models to the v2 pattern, including configuration, micro-iteration workflow, and success criteria.
* Introduced a comprehensive checklist covering all aspects of migration, such as types, field configurations, builders, presets, constants, tests, and housekeeping tasks.
* Defined rules for maintaining backwards compatibility and ensuring adherence to official guidelines and examples.

Note: I tested this prompt with migrating the [Location model](https://github.com/commercetools/test-data/pull/927). There were a few minor tweaks I made to the code and updated this prompt as a result. Upon reviewing the code changes the prompt made, I determined that one preset was no longer needed as this is now the default. I removed this one in a 2nd commit on that branch to keep what AI did separate from what I, as the developer, recognized needed to be done (although I still let Cursor make the changes based on my request).

image

I took the changes to the prompts I made with Location and tried again with the [Associate Role model](https://github.com/commercetools/test-data/pull/929). It went much faster as I got better at not trying to multi task while it did micro-iterations, and telling it not to do micro-iterations in certain spots but also where to stop. The one thing it still struggled with was knowing which props were required vs. optional in a model. Even after improving the prompt it still struggled. I wondered if there was something weird about our docs that it has a hard time reading.

### Verification Prompt (`verify-model.md`)

* Added a prompt to verify that v2 test-data models comply with the latest conventions, including file structure, builder exports, field configurations, and random presets.
* Included a checklist for validating each model against specific criteria, such as required/optional properties, compatibility layers, and test coverage.
* Provided a workflow for scanning models, reporting violations, and suggesting fixes to ensure consistency across the repository.

Note: This is actually the prompt I started with as I noticed the original attempts I made for a migration prompt were [not going well at all](https://www.dbreunig.com/2025/06/22/how-contexts-fail-and-how-to-fix-them.html). When it was discovered there were some inconsistencies between the docs and the existing v2 models (inconsistencies between models and docs, and between different model implementations), I decided to start small and see if I could use AI to help identify these inconsistencies. This prompt was used to find [this one model](https://github.com/commercetools/test-data/pull/927/files#r2201435671) that had some very small naming inconsistencies. It is also referenced in the migration prompt as a way to validate that the migration was done correctly, but there may or may not be a long term use for this one, especially if the migration prompt includes all the specific instructions that this one validates.


Sarah4VT:
Adding reviewers to this, but leaving in draft as this may or may not be something people find useful and want to use.

ByronDWall (approved):
These prompts look fine to me. I think that having them here is fine, people can use them or not, just like ai in general. For those of us who do not use cursor but have copilot, would it make sense to add a `README` or somesuch describing how you can set these prompt files to be used in VSCode?  Essentially it would just restate or point to [this 'customize AI responses in github copilot' documentation section](https://code.visualstudio.com/docs/copilot/copilot-customization#_settings)

Sarah4VT:
> These prompts look fine to me. I think that having them here is fine, people can use them or not, just like ai in general. For those of us who do not use cursor but have copilot, would it make sense to add a `README` or somesuch describing how you can set these prompt files to be used in VSCode? Essentially it would just restate or point to [this 'customize AI responses in github copilot' documentation section](https://code.visualstudio.com/docs/copilot/copilot-customization#_settings)

Honestly there is no cursor magic (that I have found yet) to having them located here. I probably should have put them in a different directory for that reason, and to be more copilot-friendly. Any suggestions for a good spot?

Sarah4VT:
Interesting
image

PR #932, opened 2025-07-15
chore: Adding discount code info model

krishhna123:
With current cart model, `discountCodes` was just array of the discount codes. But actually it should be array of `DiscountCodeInfo`. As `DiscountCodeInfo` model was not available, I have added it in this PR and also fixing cart model

### DIFF
--- a/standalone/src/models/cart/cart/discount-code-info/constants.ts
+++ b/standalone/src/models/cart/cart/discount-code-info/constants.ts
@@ -1,3 +1,3 @@
 import { TCtpDiscountCodeState } from '@/graphql-types';
 
-export const discountCodeState = TCtpDiscountCodeState;
+export const state = TCtpDiscountCodeState;
--- a/standalone/src/models/cart/cart/discount-code-info/constants.ts
+++ b/standalone/src/models/cart/cart/discount-code-info/constants.ts
@@ -1,0 +1,3 @@
+import { TCtpDiscountCodeState } from '@/graphql-types';
+
+export const discountCodeState = TCtpDiscountCodeState;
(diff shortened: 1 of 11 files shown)


standalone/src/models/cart/cart/discount-code-info/constants.ts
```diff
@@ -0,0 +1,3 @@
+import { TCtpDiscountCodeState } from '@/graphql-types';
+
+export const discountCodeState = TCtpDiscountCodeState;
```
CarlosCortizasCT, line 3: I would suggest to use a simplified name as the access to this constant will be namespaced: `DiscountInfoGraphql.constants.states`.

standalone/src/models/cart/cart/discount-code-info/index.ts
```diff
@@ -0,0 +1,16 @@
+import { RestModelBuilder, GraphqlModelBuilder } from './builders';
+import * as constants from './constants';
+import * as DiscountCodeInfoPresets from './presets';
+export * from './types';
```
CarlosCortizasCT, line 4: The types are already exported in the `index.ts` file so there's no need to export them here as well (repeated).

standalone/src/models/cart/cart/discount-code-info/constants.ts
krishhna123: update the constant name to `state`

krishhna123:
> Sorry to jump-in here but I wanted to share a couple of suggestions for consistency with other models.

No problem at all @CarlosCortizasCT , your reviews are always helpful :-)

standalone/src/models/cart/cart/discount-code-info/constants.ts
krishhna123: Done

standalone/src/models/cart/cart/discount-code-info/index.ts
krishhna123: Removed exports from model's index file as you suggested. I agree with you that once we have new implementations everywhere, it will be easy.

CarlosCortizasCT (approved):
Thanks for following-up on my feedback 🙇

PR #935, opened 2025-07-22
Add new presets to Cart models

CarlosCortizasCT:
## Description

We're introducing two new `withAllFields` presets for the `DiscountedLisItemPortion` and `DiscountedLineItemPrice` that can be used to generate objects for those models with all the fields populated.

They both accept some params which allow to provide some key values used for the nested models.

Here are some examples on how they can be used:
```ts
import {
  DiscountedLineItemPriceGraphql,
  DiscountedLineItemPortionGraphql,
} from '@commercetools/composable-commerce-test-data/cart';

// DiscountedLineItemPrice

// When no param is provided we will use these values
//  - currencyCode: 'EUR'
//  - target: 'lineItems'
//  - discountId: 'cart-discount-id'
const discountedLineItemPrice = DiscountedLineItemPriceGraphql.presets.withAllFields();

// With some custom params
const discountedLineItemPriceCustomized = DiscountedLineItemPriceGraphql.presets.withAllFields({
  currencyCode: 'USD',
});


// DiscountedLineItemPortionGraphql

// When no param is provided we will use these values
//  - currencyCode: 'EUR'
//  - discountId: 'cart-discount-id'
const discountedLineItemPortion = DiscountedLineItemPortionGraphql.presets.withAllFields();

// With some custom params
const discountedLineItemPriceCustomized = DiscountedLineItemPriceGraphql.presets.withAllFields({
  discountId: 'discount-XZV',
});
```

We've also updated the `withAllFields` preset in the `LineItem` test data model so we make sure the `variant` property value that is generated with one boolean attribute filled in its attributes list property.

### DIFF
--- a/standalone/src/models/cart/cart/line-item/constants.ts
+++ b/standalone/src/models/cart/cart/line-item/constants.ts
@@ -1,16 +1,11 @@
-export const inventoryMode = {
-  TrackOnly: 'TrackOnly',
-  ReserveOnOrder: 'ReserveOnOrder',
-  None: 'None',
-} as const;
+import {
+  TCtpInventoryMode,
+  TCtpLineItemMode,
+  TCtpLineItemPriceMode,
+} from '@/graphql-types';
 
-export const priceMode = {
-  Platform: 'Platform',
-  ExternalPrice: 'ExternalPrice',
-  ExternalTotal: 'ExternalTotal',
-} as const;
+export const inventoryMode = TCtpInventoryMode;
 
-export const lineItemMode = {
-  Standard: 'Standard',
-  GiftLineItem: 'GiftLineItem',
-} as const;
+export const priceMode = TCtpLineItemPriceMode;
+
+export const lineItemMode = TCtpLineItemMode;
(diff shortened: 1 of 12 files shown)


standalone/src/models/cart/cart/line-item/constants.ts
```diff
@@ -1,16 +1,11 @@
-export const inventoryMode = {
-  TrackOnly: 'TrackOnly',
-  ReserveOnOrder: 'ReserveOnOrder',
-  None: 'None',
-} as const;
+import {
```
CarlosCortizasCT, line 1: Instead of using raw string values, we now export the GraphQL generated TS enums so we can populate the properties with the actual expected types.
REST model can also use the string values of each enum item.

standalone/src/models/commons/reference/types.ts
```diff
@@ -1,22 +1,6 @@
 import type { TBuilder } from '@/core';
 import { TCtpReference } from '@/graphql-types';
 
-// export interface BusinessUnitReference {
```
CarlosCortizasCT, line 4: Unrelated but I noticed this useless comment block and took the opportunity to remove it.

PR #939, opened 2025-07-23
Price test data model migration

CarlosCortizasCT:
## Description

The main goal of this PR is to migrate the `Price` test data model to the new implementation patterns.

While doing so, I noticed its GraphQL draft version needed a new model (`BaseMoneyDraft`) which I also implemented.

Finally, I found some models which should be using the new migrated `Money` model types, so I took the opportunity to update them as well.


standalone/src/models/cart/cart/discounted-line-item-portion/fields-config.ts
```diff
@@ -10,15 +15,15 @@ export const restFieldsConfig: TModelFieldsConfig<TDiscountedLineItemPortionRest
   {
     fields: {
       discount: fake(() => ReferenceRest.presets.cartDiscountReference()),
-      discountedAmount: fake(() => Money.random()),
+      discountedAmount: fake(() => MoneyRest.random()),
```
CarlosCortizasCT, line 18: Not related to the `Price` migration but I noticed we also need to start using the new `Money` models so I took the opportunity to update some of the models I identified.

standalone/src/models/commons/base-money/base-money-draft/fields-config.ts
```diff
@@ -0,0 +1,11 @@
+import { type TModelFieldsConfig } from '@/core';
+import type { TBaseMoneyDraftGraphql } from '../types';
+
+// This model only exists in the GraphQL API
```
CarlosCortizasCT, line 4: According to the GraphQL API, the `PriceDraft` entity `value` field has this type:
```ts
export type TCtpBaseMoneyInput = {
  centPrecision?: InputMaybe<TCtpMoneyInput>;
  highPrecision?: InputMaybe<TCtpHighPrecisionMoneyInput>;
};
```

It does not seem to have something similar for the REST version as the `value` field is defined like this for that one:
```ts
export type _Money = Money | TypedMoney | TypedMoneyDraft;
```

standalone/src/models/commons/price/price-draft/fields-config.ts
```diff
@@ -0,0 +1,46 @@
...
+  fields: {
+    ...commonFieldsConfig.fields,
+    value: fake(() => MoneyDraftRest.random()),
+  },
+};
+
+export const graphqlFieldsConfig: TModelFieldsConfig<TPriceDraftGraphql> = {
+  fields: {
+    ...commonFieldsConfig.fields,
+    value: fake(() => BaseMoneyDraftGraphql.presets.withAllFields()),
```
CarlosCortizasCT, line 33: @commercetools/priceless-team-fe It would be great if you could let me know if I understood this one correctly or if it should be any different 🙏.

jmcreasman (approved):
👍

ragafus (commented):
A question out of the scope of this PR...
The existing `Price` model in `commons` is really one type of price. In Priceless we have product prices (AKA embedded prices, what `Price` is modeling here) and standalone prices (what `StandalonePrice` is modeling in this repo). So, naming one of the types as `Price` and to have it in a `commons` package is confusing. If we wanted to move it and rename it in the future, when should Priceless introduce that breaking change?

standalone/src/models/commons/base-money/base-money-draft/fields-config.ts
```diff
@@ -0,0 +1,11 @@
...
+
+export const graphqlFieldsConfig: TModelFieldsConfig<TBaseMoneyDraftGraphql> = {
+  fields: {
+    centPrecision: null,
+    highPrecision: null,
```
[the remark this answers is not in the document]
CarlosCortizasCT: In the fields configuration we only populate the mandatory fields and, according to the TS types, both are optional.

The way I see it, perhaps it makes sense to keep this as it is and change the preset. Instead of having the `withAllFields`, we could have two different ones (`withCentPrecision` and `withHighPrecision`) that could be used by consumers to get a populated model.

Another option is to populate here one of both but it would be your team the one deciding which one to populate.

standalone/src/models/commons/price/price-draft/fields-config.ts
CarlosCortizasCT: Based on my previous comment, perhaps we could use one of the new proposed presets here and populate the value with a cent or high precision money.

For what I've seen in other models, usually we're populating with cents.

What do you think?

CarlosCortizasCT:
> A question out of the scope of this PR... The existing `Price` model in `commons` is really one type of price. In Priceless we have product prices (AKA embedded prices, what `Price` is modeling here) and standalone prices (what `StandalonePrice` is modeling in this repo). So, naming one of the types as `Price` and to have it in a `commons` package is confusing. If we wanted to move it and rename it in the future, when should Priceless introduce that breaking change?

I would say that if you can wait until the FE Chapter finalizes the current migration work in the MC monorepo that would help, but it's up to you when you want to work on that.
You will just need to bear in mind how you want to lay out the components in the directories structure, which entry points you want to update and plan for how consumers would need to be updated.

standalone/src/models/commons/base-money/base-money-draft/fields-config.ts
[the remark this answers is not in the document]
ragafus: > The way I see it, perhaps it makes sense to keep this as it is and change the preset. Instead of having the `withAllFields`, we could have two different ones (`withCentPrecision` and `withHighPrecision`) that could be used by consumers to get a populated model.

It makes totally sense 👍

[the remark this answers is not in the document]
CarlosCortizasCT: Refactored in this commit: 124e06ccdeab57616e4ba0668acba1a2fbd5fa32

ragafus (approved):

PR #942, opened 2025-08-01
New ImageDimensions test data model

CarlosCortizasCT:
## Description

I started working on the migration of the `Image` test data model and realized we don't currently have an `ImageDimensions` model which the former relies on.

I decided to first open this PR only adding the new `ImageDimensions` test data model so we can split the review of the whole work needed.


dogayuksel (approved):

jmcreasman (approved):
👍

PR #955, opened 2025-09-02
FEC-240: migrate shipping method models to new model

LucasGrahn:
This PR is needed to make the migration of the e2e tests to the new test data model inside the project settings app from the MC possible.

There seems to be troubles when the old model is used with the new version that is why the adding of the ShippingRatePriceTier model provoked the changes of related/parent models.

**Added**
ShippingRatePriceTier [[Ref](https://docs.commercetools.com/api/projects/shippingMethods#shippingratepricetier)]

**Migrated**
ShippingRate [[Ref](https://docs.commercetools.com/api/projects/shippingMethods#shippingrate)]
ShippingRateDraft [[Ref](https://docs.commercetools.com/api/projects/shippingMethods#shippingratedraft)]

**Planned**
ShippingMethod
ZoneRate
ZoneRateDraft

PR #969, opened 2025-10-07
FEC-240: Migrate zone-rate models  

nima-ct:
This PR is the second step of splitting work originally started in [PR #955](https://github.com/commercetools/test-data/pull/955).

It focuses on migrating the `ZoneRate` and `ZoneRateDraft` models to the new implementation patterns. This migration is required to continue enabling the Project Settings app e2e test migration to the new test data model in the MC.


.changeset/quick-bugs-bow.md
```diff
@@ -0,0 +1,31 @@
+---
+'@commercetools/composable-commerce-test-data': patch
+---
+
+## Migrated Zone Rate Model
+
+The zone rate models have been migrated to the new model structure:
```
CarlosCortizasCT, line 7: I would suggest to use [the template](https://github.com/commercetools/test-data/blob/main/docs/guidelines/writing-changesets.md#if-you-are-migrating-an-existing-model) we have in the guidelines so we let know consumers that we still have the `ZoneRate` model but it's now deprecated.

standalone/src/models/shipping-method/zone-rate/zone-rate-draft/presets/change-history-data/with-usd-shipping-rate.ts
```diff
@@ -1,16 +1,37 @@
...
+const populatePreset = <
+  TModel extends TZoneRateDraftGraphql | TZoneRateDraftRest | TZoneRateDraft,
+>(
+  builder: TBuilder<TModel>
+) => {
+  return builder
     .zone(KeyReferenceDraft.presets.zone().key('e2e-us-zone'))
     .shippingRates([
-      ShippingRateDraft.presets.changeHistoryData.withUsdCurrency(),
+      ShippingRateDraftRest.presets.changeHistoryData.withUsdCurrency(),
```
CarlosCortizasCT, line 26: We can't use the rest model for all the presets as each type needs a different one.
The shippingRates array should be a function parameter each preset type should provide.

Something like this:
```ts
export const graphqlPreset = (): TBuilder<TZoneRateDraftGraphql> =>
  populatePreset(GraphqlModelBuilder(), [ShippingRateDraftGraphql.presets.changeHistoryData.withUsdCurrency()]);
```

standalone/src/models/shipping-method/zone-rate/zone-rate-draft/fields-config.ts
```diff
@@ -0,0 +1,21 @@
+import { fake, type TModelFieldsConfig } from '@/core';
+import { KeyReferenceDraft } from '@/models/commons';
+import { ShippingRateDraftRest } from '../../shipping-rate/shipping-rate-draft';
+import type { TZoneRateDraftRest, TZoneRateDraftGraphql } from '../types';
+
```
CarlosCortizasCT, line 5: It would help to leave a comment here with the link to this entity's docs:
```
// https://docs.commercetools.com/api/projects/shippingMethods#zoneratedraft
```

standalone/src/models/shipping-method/zone-rate/zone-rate-draft/fields-config.ts
```diff
@@ -0,0 +1,21 @@
...
+const commonFieldsConfig = {
+  zone: fake(() => KeyReferenceDraft.presets.zone()),
+  shippingRates: fake(() => [ShippingRateDraftRest.random()]),
```
CarlosCortizasCT, line 8: This can't be populated in the common fields as each fields config should use a different nested model: `ShippingRateDraftRest | ShippingRateDraftGraphql`.

standalone/src/models/shipping-method/zone-rate/builders.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+        }),
+      ]),
+    })
+  );
+}
+
+function validateGraphqlModel(model: TZoneRateGraphql) {
+  const zoneId = model.zone?.id;
+
+  expect(model).toEqual(
```
CarlosCortizasCT, line 28: When validating nested objects, it's enough to only check the `__typename` property. If that value is correct, we can assume the object is the one we're expecting.

standalone/src/models/shipping-method/zone-rate/fields-config.ts
```diff
@@ -0,0 +1,39 @@
...
+  KeyReference,
+  ReferenceGraphql,
+  TReferenceGraphql,
+} from '@/models/commons';
+import { Zone } from '@/models/zone';
+import { ShippingRateRest } from '../shipping-rate';
+import type { TZoneRateRest, TZoneRateGraphql } from './types';
+
+const commonFieldsConfig = {
+  shippingRates: fake(() => [ShippingRateRest.random()]),
```
CarlosCortizasCT, line 12: This can't be populated in the common fields as each fields config should use its corresponding nested model: `ShippingRateRest | ShippingRateGraphql`

.changeset/quick-bugs-bow.md
nima-ct: Addressed here: https://github.com/commercetools/test-data/pull/969/commits/8f3d5faf5d2b5ce9916f43d2a4b6606f363fb176 + https://github.com/commercetools/test-data/pull/969/commits/a5a27c7357a783e80ca084a790663f276094a739

standalone/src/models/shipping-method/zone-rate/zone-rate-draft/presets/change-history-data/with-usd-shipping-rate.ts
nima-ct: Addressed here: https://github.com/commercetools/test-data/pull/969/commits/596a289abefb536ddd98b6599c1685d5bd018599

CarlosCortizasCT: We shouldn't be using `build()` calls as part of the properties definition for a model so I took the opportunity to update this preset in this commit: e6cbb1796e8ac64478be86656886038924016525

jocelyn-dunkley (approved):

PR #970, opened 2025-10-07
Fix AttributeGroupDraft test-data model

CarlosCortizasCT:
## Description

We've fixed an error in the `AttributeGroupDraft` test-data model as its name was not populated with a draft model.

Also, we've updated the entry point of the `@commercetools/composable-commerce-test-data/attribute-group` package entry point so it also exports the `AttributeReference` model.
It can be used like this:
```ts
import {
  AttributeReferenceGraphql,
  AttributeReferenceRest,
} from '@commercetools/composable-commerce-test-data/attribute-group';

const graphqlModel = AttributeReferenceGraphql.random().build();
const restModel = AttributeReferenceRest.random().build();
```

I want to also migrate the `AttributeGroup` test-data model as a follow-up, but on it's own PR.

### DIFF
--- a/standalone/src/models/attribute-group/attribute-group-draft/fields-config.ts
+++ b/standalone/src/models/attribute-group/attribute-group-draft/fields-config.ts
@@ -1,5 +1,5 @@
 import { fake, type TModelFieldsConfig } from '@/core';
-import { LocalizedString } from '@/models/commons';
+import { LocalizedStringDraft } from '@/models/commons';
 import {
   AttributeReferenceRest,
   AttributeReferenceGraphql,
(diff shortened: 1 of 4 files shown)


standalone/src/models/attribute-group/attribute-group-draft/fields-config.ts
```diff
@@ -14,7 +14,7 @@ import type {
 const commonFieldsConfig = {
   key: null,
   description: null,
-  name: fake(() => LocalizedString.random()),
+  name: fake(() => LocalizedStringDraft.random()),
```
CarlosCortizasCT, line 17: This made the resulting object to contain a `__typename` property which is problematic for consumers.

standalone/src/models/attribute-group/index.ts
```diff
@@ -5,3 +5,4 @@ export * as presets from './presets';
 export * from './types';
 
 export * from './attribute-group-draft';
+export * from './attribute-reference';
```
CarlosCortizasCT, line 8: We also export this model so consumers can use it in case they need to provide some custom values to the `AttributeGroupDraft`.

nima-ct (approved):

PR #971, opened 2025-10-07
Migrate AttributeGroup test-data model to the new implementation patterns

CarlosCortizasCT:
## Description

_TBD_


standalone/src/models/attribute-group/attribute-group/fields-config.ts
```diff
@@ -0,0 +1,48 @@
...
+
+const commonFieldsConfig = {
+  attributes: fake(() => []),
+  createdAt: fake(getOlderDate),
+  createdBy: fake(() => ClientLogging.random()),
+  description: null,
+  id: fake((f) => f.string.uuid()),
+  key: null,
+  lastModifiedAt: fake(getNewerDate),
+  lastModifiedBy: fake(() => ClientLogging.random()),
```
nima-ct, line 18: `createdBy` and `lastModifiedBy` are optional fields in the specs. Is there a specific reason they are being set here?

CarlosCortizasCT: I populated them for consistency because every other test-data model in the repo that has these fields are populating them

nima-ct: Good to know! I’ll include them again in my current migration as well

nima-ct (approved):

